PR #82, opened 2026-04-08
fix: handle date fields returned as strings by redminelib

mihajlovicjj:
## Description
Fix AttributeError crash when redminelib returns date fields as pre-formatted
strings instead of datetime objects.

## Related Issue
Reported in production: `list_redmine_projects` fails with
`'str' object has no attribute 'isoformat'` on certain Redmine server
configurations (non-UTC timezone / different Redmine versions).

## Changes Made
- Added `_safe_isoformat()` helper that passes strings through unchanged and
  only calls `.isoformat()` on actual date/datetime instances
- Replaced all 10 bare `.isoformat()` calls on Redmine date fields
  (`created_on`, `updated_on`) across issues, journals, attachments, versions,
  projects, and time entries
- Left locally-constructed datetime `.isoformat()` calls untouched (they are
  always real datetime objects)

## Checklist
- [x] Code follows style guidelines
- [x] Tests added
- [x] CHANGELOG updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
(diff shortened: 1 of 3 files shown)


jztan (approved):
LGTM — the fix is correct and well-structured. We'll follow up with additional test coverage for `_issue_to_dict_selective` and `list_redmine_projects` in a separate PR.

PR #95, opened 2026-04-21
feat: add inline attachment retrieval modes

sebastianelsner:
I am using redmine mcp in stdio mode. Why? because I want each codex instance to start it's own redmine mcp, so I do not need to handle the start of it manually and I can't run a http mcp with a fixed port multiple times. Since the http server is not running, downloading redmine attachments does not work. So here is the fix.

## Summary
- add get_redmine_attachment_content for stdio-safe inline attachment retrieval using MCP-native content blocks
- keep get_redmine_attachment_download_url for HTTP deployments and add ATTACHMENT_TOOL_MODE to publish only the attachment tool or tools appropriate for the deployment
- document the new attachment modes and cover them with unit, security, and integration tests

## What this changes
This branch adds a second attachment retrieval path alongside the existing URL-based flow.

get_redmine_attachment_content streams the attachment into temporary server storage under a hard byte cap and returns it directly in the MCP response as:
- ImageContent for images
- AudioContent for audio
- TextContent for text-like files
- EmbeddedResource blobs for other binary files

The inline path checks Redmine metadata first, aborts streaming as soon as the configured size limit is exceeded, sanitizes user-controlled filenames, and removes temporary files after the response is built.

The branch also introduces ATTACHMENT_TOOL_MODE with inline, url, and both modes so operators can expose only the attachment tool that matches how the server is deployed.

## Why this is a relevant fix
The existing get_redmine_attachment_download_url flow assumes the MCP client can reach the server /files/{id} HTTP route. That works for HTTP deployments, but it breaks down for stdio-based MCP clients because there is no reliable localhost side channel for the client to fetch the file.

This change fixes that deployment mismatch by letting stdio clients receive attachments directly in the MCP response, in the content shape the protocol already expects, without forcing binary data through ad hoc JSON fields or requiring an extra HTTP hop.

The tool-mode gate is relevant because it prevents servers from advertising an attachment workflow that is unusable in the current transport mode, while still preserving the existing URL-based flow for deployments where it is the right choice.

## Validation
- uv run --extra test pytest -q tests/test_redmine_handler.py -k "attachment or AttachmentToolRegistration"
- uv run --extra test pytest -q tests/test_security_validation.py


jztan:
Hi @sebastianelsner, thanks for this contribution — the inline attachment retrieval approach is exactly the right fix for stdio deployments.

One housekeeping item: the PR was originally targeting `master`, which I've corrected to `develop` (our active development branch). This introduced merge conflicts in three files:

- `README.md`
- `docs/tool-reference.md`
- `src/redmine_mcp_server/redmine_handler.py` (imports only)

Could you rebase your branch onto `develop` and force-push? The conflicts are minor — `develop` has added several new tools since `master`, so it's mostly a matter of slotting your changes into the updated file structure.

```bash
git fetch upstream
git rebase upstream/develop
# resolve conflicts, then:
git push --force-with-lease
```

The `redmine_handler.py` conflict is just two import blocks that need to be merged. The `README.md` and `docs/tool-reference.md` conflicts are in the File Operations sections where new tools were added on `develop` — your new tool entries just need to be placed alongside them.

Let me know if you run into any issues.

sebastianelsner:
Will do!

jztan:
Hi @sebastianelsner, thank you for identifying the stdio attachment gap and putting together a detailed fix. The problem statement is exactly right.

Since this PR was opened, we shipped `get_redmine_attachment` in the v2.0 refactor, which I think covers your use case in a way that avoids the inline content trade-offs:

**What we shipped**

`get_redmine_attachment` is a unified tool that detects the deployment mode automatically from `PUBLIC_HOST`/`SERVER_HOST` at call time:

- **stdio mode** (no external public host configured): returns `file_path` -- an absolute local path the model can pass directly to Claude Code's `Read` tool or pdf-mcp, with no HTTP round-trip needed
- **HTTP mode** (external hostname configured): returns `uri` -- the existing time-limited HTTP URL

No `ATTACHMENT_TOOL_MODE` configuration is required. The caller gets a usable reference either way without knowing which mode is active.

**Why `file_path` instead of inline content**

Returning inline content (ImageContent / AudioContent / base64 blobs) works, but it comes with real costs that are hard to opt out of:

- All inline content is base64-encoded into the JSON-RPC frame, inflating byte count ~33% and burning context window tokens on every call
- Text and binary content lands directly in the LLM context, expanding the prompt injection surface beyond what `<insecure-content>` tags alone can contain
- Clients that do not support a given content type (AudioContent especially) get a silent failure or a confusing blob

The `file_path` approach sidesteps all of this: the file lives on disk, the model decides whether and how to read it, and nothing goes into the context window unless the model explicitly requests it.

**Current state**

The codebase has also been restructured from the `redmine_handler.py` monolith into a `tools/` package, so this PR would need significant rework to apply cleanly regardless.

Closing this in favour of the shipped implementation. If you run into cases where `file_path` does not cover your workflow -- for example, an environment where the model cannot access the local filesystem -- please open a new issue and we can discuss inline content as a follow-up phase.

Thanks again for the contribution.

sebastianelsner:
Hi! Yes, this is the right solution, I was thinking about it as well when implementing the original draft and again in recent times when using the mcp. My solution seemed "clumsy" and this was also why I was hesitant to merge.
We have the security policy that every - lets call it - "AI tool" needs to run in it's own isolated low privilege container with the least amount of paths mapped in (we use Apptainer for this.) So usually the Apptainer container does not share a filesystem, so I need to map the external path to the same path in the container and your new solution should work. Maybe you see why the inline content appealed more to me initially. Thank you for the work, I will report back once I get the time to check out the newest version!

Issue #118, opened 2026-05-15
Wiki attachments missing content_url breaks cross-tool symmetry with get_redmine_issue


jztan:
## Problem

\`get_redmine_issue(issue_id=N, include_attachments=True)\` returns attachments with a \`content_url\` field. \`manage_redmine_wiki_page(action=\"get\", include_attachments=True)\` returns attachments **without** \`content_url\`:

\`\`\`python
# get_redmine_issue:
{\"id\": 3, \"filename\": \"spec.pdf\", \"content_url\": \"...\", \"author\": {...}, ...}

# manage_redmine_wiki_page (action=\"get\"):
{\"id\": 3, \"filename\": \"spec.pdf\", \"filesize\": ..., \"content_type\": \"...\", ...}
#                                       ^^^^^^^^^^^ missing
\`\`\`

The cross-tool asymmetry costs an agent a turn: \"does this attachment have a URL? let me check the other tool's response shape.\"

## Proposal

In \`_wiki_page_to_dict\` (\`src/redmine_mcp_server/tools/wiki.py\`), include \`content_url\` in the per-attachment dict, route it through \`_rewrite_to_public_url\` (so subpath / public-host handling stays consistent with #110), and add \`author\` via \`_named_ref\` so the shape matches \`_attachments_to_list\`.

Ideally, both call sites would share a single \`_attachment_to_dict\` helper instead of duplicating the per-attachment construction. Mention this in the PR.

## Tests

- Assert \`manage_redmine_wiki_page(action=\"get\", include_attachments=True)\` returns \`content_url\` on every attachment.
- Assert \`_rewrite_to_public_url\` is applied (mirror the end-to-end tests in \`tests/test_content_url_rewriting.py\`).

## Out of scope

- Removing the in-line per-attachment dict in favor of a shared helper is a nice refactor but optional; the bug fix is the missing field.

Reference: surfaced during verification of #110.

jztan:
**Unblocked by #109:** the wrap policy is now settled (filenames + names verbatim, free-text wrapped). Collapsing \`_attachments_to_list\` (tools/issues.py:354) and the inline wiki-page attachment construction (tools/wiki.py:64-79) into a shared \`_attachment_to_dict\` helper becomes a near-mechanical refactor -- the new helper just needs to mirror what \`_attachments_to_list\` does today.

See \`tests/test_wrap_policy.py\` for the policy fixtures the new helper should keep passing.

jztan:
[...]

Issue #117, opened 2026-05-15
Consolidate list/search error envelopes via _format_list_error helper


jztan:
## Problem

Error responses from \`list_*\` and \`search_*\` tools are not consistent. Three shapes observed during verification of #108:

1. \`{\"error\": \"...\", \"code\": \"...\", \"hint\": \"...\"}\` (from \`_handle_redmine_error\` — the target shape)
2. \`[{\"error\": \"...\"}]\` (top-level array wrapping a single-item dict)
3. Bare strings leaked from FastMCP's output validation (now fixed for input errors via #108's middleware)

\`CleanValidationErrorMiddleware\` only catches argument-validation errors. It cannot unify the post-execution error paths because each list tool returns its own shape directly.

## Proposal

Add \`_format_list_error(exc, operation, context)\` in \`_errors.py\`, mirroring \`_handle_redmine_error\` but returning the dict envelope wrapped or shaped appropriately for tools that declare \`Union[List, Dict]\` return types. Every \`list_*\` and \`search_*\` tool's \`except\` block routes through this helper.

## Audit step (do this first)

Grep for \`return [{\"error\":\`, \`return [\` inside \`except\` blocks, and any other ad-hoc error shapes in the tool layer. Capture the full inventory in a comment on this issue before writing the helper, so the helper is shaped by real call sites rather than a guess.

Likely sites: \`list_redmine_issues\`, \`search_redmine_issues\`, \`search_entire_redmine\`, \`list_time_entries\`, \`list_files\`, \`list_redmine_versions\`, \`list_subtasks\`, \`list_issue_relations\`, \`list_redmine_users\`, \`list_redmine_roles\`, \`list_redmine_trackers\`, \`list_redmine_issue_statuses\`, \`list_redmine_issue_priorities\`, \`list_redmine_queries\`, \`list_issue_categories\`.

## Tests

A single parametrized test that fires a controlled exception inside each tool's \`except\` block and asserts the envelope shape matches the standardized form.

## Out of scope

- The middleware-level \`INVALID_ARGUMENTS\` path (already standardized in #108).
- Non-list tools (\`get_*\`, \`create_*\`, etc.) — those already go through \`_handle_redmine_error\`.

Reference: see #108's deferral notes for context.

jztan:
Fixed in $(git rev-parse --short HEAD) (on \`develop\`).

[...]

Issue #116, opened 2026-05-15
Tighten user-ID-style params: int | Literal["me"] instead of str | int


jztan:
## Problem

Several tools accept user-ID-style parameters typed as \`Union[int, str]\` so callers can pass either a numeric ID or the Redmine sentinel \`\"me\"\`. The wide \`str\` half silently accepts arbitrary garbage:

- \`list_redmine_issues(assigned_to_id=\"garbage\")\` returns \`[]\` instead of an error.
- A model retrying on empty results can't tell whether the filter is wrong or whether no issues actually match.

## Proposal

Tighten the type to \`Optional[Union[int, Literal[\"me\"]]]\` at every site where the \"me\" sentinel is the only valid string value. FastMCP will surface the literal as an enum in the JSON schema, and \`CleanValidationErrorMiddleware\` (#108) will reject other strings with the standardized \`INVALID_ARGUMENTS\` envelope.

## Inventory work (start here)

Grep for \`assigned_to_id\` and \`user_id\` across the tool surface and audit which accept the \"me\" sentinel:

- \`list_redmine_issues.assigned_to_id\` — confirmed accepts \`\"me\"\`
- \`list_time_entries.user_id\` — likely
- \`manage_time_entry.user_id\` (for the create-on-behalf path) — likely
- \`import_time_entries\` per-entry \`user_id\` — likely
- Anywhere else the same pattern shows up

Document each in the issue before patching so the change lands as one clean PR.

## Tests

Each tightened param needs:
1. A schema test that asserts \`enum\` membership includes \`\"me\"\` (analogous to \`tests/test_manage_action_schema.py\`).
2. A runtime test that an invalid string is rejected with the \`INVALID_ARGUMENTS\` envelope (not silently passed through to Redmine).

## Out of scope

- Tools where \`str\` is genuinely free-form (project identifiers, search queries).

Reference: see #108's deferral notes for context.

jztan:
Fixed in c41a300 (on \`develop\`).

**Type change:** \`list_redmine_issues.assigned_to_id\` and \`list_time_entries.user_id\` are now \`Optional[Union[int, Literal[\"me\"]]]\` (was \`Optional[Union[int, str]]\`). Schema renders the string branch as a \`const: \"me\"\` enum.

**Runtime:** \`assigned_to_id=\"notmeortheid\"\` now produces the standardized \`INVALID_ARGUMENTS\` envelope (via #108):

\`\`\`json
{
[...]

Issue #115, opened 2026-05-15
Gate cleanup_attachment_files out of the agent-facing surface


jztan:
## Problem

\`cleanup_attachment_files\` is an admin/cron task — there's no realistic agent workflow that should be calling it. Exposing it as a tool adds discovery noise and a foot-gun.

## Proposal

Gate behind a flag (e.g., \`EXPOSE_ADMIN_TOOLS=true\`) or move out of the agent-facing surface entirely. The background cleanup task already handles this; the tool is mostly an ops escape hatch.

Source: LLM-driven evaluation, 2026-05-15.

jztan:
Fixed in $(git rev-parse --short HEAD) (on \`develop\`).

\`cleanup_attachment_files\` is no longer registered on the default MCP surface. The background cleanup task continues to run on the \`CLEANUP_INTERVAL_MINUTES\` schedule regardless -- gating only affects whether the function is reachable via \`tools/list\` and \`call_tool\`.

**Opt-in:** \`REDMINE_MCP_EXPOSE_ADMIN_TOOLS=true\`

The function remains importable for internal callers (tests, scripts, automation) regardless of the flag; only the MCP surface registration is conditional.

**Drift guard:** \`tests/test_admin_tools_gating.py\` (6 tests) covers the helper's truthy/falsy parsing, the default-off path, the opt-in path via \`importlib.reload\`, and that the function remains importable when gated off.

PR #121, opened 2026-05-15
fix: restrict user id sentinels to me

nanookclaw:
## What

Tightens the user-ID-style filters that support Redmine's `"me"` sentinel so they accept only numeric IDs or the literal `"me"`, instead of any arbitrary string.

This updates:

- `list_redmine_issues.assigned_to_id`
- `list_time_entries.user_id`

`manage_time_entry.user_id` and bulk import per-entry `user_id` remain numeric-only because those paths log time on behalf of a user and already validate positive integer IDs.

## Why

With `Union[int, str]`, callers could pass values like `"garbage"`; FastMCP would treat them as schema-valid and the value would be forwarded to Redmine, often returning an empty result that looks like a valid no-match query. Using `Literal["me"]` lets the tool schema expose the sentinel as an enum and lets the existing validation middleware reject other strings with `INVALID_ARGUMENTS`.

## Validation

- `uv run --extra test -- python -m pytest tests/test_user_id_literal_me.py -q`
- `uv run --extra test -- python -m pytest tests/test_list_redmine_issues.py tests/test_time_entries.py tests/test_time_tracking_tools.py tests/test_tool_error_middleware.py tests/test_manage_action_schema.py tests/test_limit_offset_schema.py -q`
- `uv run --extra dev -- python -m flake8 src/redmine_mcp_server/tools/issues.py src/redmine_mcp_server/tools/time_tracking.py tests/test_user_id_literal_me.py`
- `uv run --extra dev -- python -m black --check src/redmine_mcp_server/tools/issues.py src/redmine_mcp_server/tools/time_tracking.py tests/test_user_id_literal_me.py`

### DIFF
--- a/src/redmine_mcp_server/tools/issues.py
+++ b/src/redmine_mcp_server/tools/issues.py
@@ -570,7 +570,7 @@ async def list_redmine_issues(
     project_id: Optional[Union[int, str]] = None,
     status_id: Optional[Union[int, Literal["open", "closed", "*"]]] = None,
     tracker_id: Optional[int] = None,
-    assigned_to_id: Optional[Union[int, str]] = None,
+    assigned_to_id: Optional[Union[int, Literal["me"]]] = None,
     priority_id: Optional[int] = None,
     fixed_version_id: Optional[int] = None,
     sort: Optional[str] = None,
--- a/src/redmine_mcp_server/tools/time_tracking.py
+++ b/src/redmine_mcp_server/tools/time_tracking.py
@@ -57,7 +57,7 @@ def _time_entry_to_dict(time_entry: Any) -> Dict[str, Any]:
 async def list_time_entries(
     project_id: Optional[Union[str, int]] = None,
     issue_id: Optional[int] = None,
-    user_id: Optional[Union[str, int]] = None,
+    user_id: Optional[Union[int, Literal["me"]]] = None,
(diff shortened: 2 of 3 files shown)


jztan:
Thanks for filing this @nanookclaw -- closing as a duplicate of #116, which shipped in c41a300 on \`develop\` right before your issue went up.

The fix in #116 makes exactly the change you describe:

- \`list_redmine_issues.assigned_to_id\` and \`list_time_entries.user_id\` are now \`Optional[Union[int, Literal[\"me\"]]]\`
- \`manage_time_entry.user_id\` and the bulk-import per-entry \`user_id\` left as numeric-only (same audit conclusion as yours -- those paths log time on behalf of a user and the existing positive-integer validator is sufficient)
- Out-of-set strings are rejected at the FastMCP boundary by \`CleanValidationErrorMiddleware\` (#108) with an \`INVALID_ARGUMENTS\` envelope that lists both accepted shapes

Drift guard for both branches lives in \`tests/test_user_id_typing_schema.py\`.

Sorry for the collision -- this was on the active backlog from the LLM eval and got picked up earlier today. If you'd noticed the existing #116 first I would've happily handed it off; please don't let this discourage you from picking up other issues. The \`good first issue\` label on the remaining backlog (#114, #115) is current, and #117 / #118 are open and well-scoped if you want a larger surface.

PR #135, opened 2026-05-23
Serve OAuth AS metadata at MCP path

aadnehovda:
## Summary

This PR narrows the MCP OAuth Authorization Server Metadata route from the root path to the MCP resource path:

- `/.well-known/oauth-authorization-server` is removed
- `/.well-known/oauth-authorization-server/mcp` becomes the served AS metadata path
- docs, changelog, and OAuth discovery tests are updated for the breaking path change

This keeps the AS metadata path aligned with clients that derive RFC 8414 metadata from the MCP resource path. It also avoids the MCP server claiming the root AS metadata location when it is co-hosted with another app that may own that root discovery document.

## Out of scope

Discovery mirroring / DCR pass-through was split out of this PR. This branch keeps the existing static Doorkeeper metadata behavior and only changes the served path.

## Verification

```bash
UV_CACHE_DIR=/tmp/uv-cache-redmine-mcp uv run --extra dev pytest tests/test_oauth_auth.py tests/test_oauth_discovery.py tests/test_auth_factory.py
# 18 passed

UV_CACHE_DIR=/tmp/uv-cache-redmine-mcp uv run --extra dev black --check --fast src/redmine_mcp_server/main.py tests/test_oauth_discovery.py tests/test_oauth_auth.py
# 3 files would be left unchanged
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,9 +15,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **Two new required env vars in OAuth mode**: `REDMINE_INTROSPECT_CLIENT_ID` and `REDMINE_INTROSPECT_CLIENT_SECRET`. Operators register a confidential OAuth client in Redmine and patch Doorkeeper's `allow_token_introspection` block (stock Redmine ships with `allow_token_introspection false`). See [`docs/oauth-setup.md`](docs/oauth-setup.md) Step 2 for the walkthrough. Server fails fast at startup if either env var is missing.
 - **Discovery path aliases dropped.** Only the canonical paths remain:
   - `GET /.well-known/oauth-protected-resource/mcp` (RFC 9728 §3.1 suffix-scoped, mounted natively by `RemoteAuthProvider`)
-  - `GET /.well-known/oauth-authorization-server` (canonical root, kept as `custom_route` mirror of Redmine's Doorkeeper AS metadata)
+  - `GET /.well-known/oauth-authorization-server/mcp` (path-scoped RFC 8414 form, kept as `custom_route` mirror of Redmine's Doorkeeper AS metadata)
 
-  These previously-served paths now return 404: `/.well-known/oauth-protected-resource` (root), `/mcp/.well-known/oauth-protected-resource` (prefix), `/.well-known/oauth-authorization-server/mcp` (suffix), `/mcp/.well-known/oauth-authorization-server` (prefix). Clients should follow `WWW-Authenticate: Bearer resource_metadata="..."` headers from 401 responses for RFC 9728 §5.3 compliant discovery.
+  These previously-served paths now return 404: `/.well-known/oauth-protected-resource` (root), `/mcp/.well-known/oauth-protected-resource` (prefix), `/.well-known/oauth-authorization-server` (root), `/mcp/.well-known/oauth-authorization-server` (prefix). Clients should follow `WWW-Authenticate: Bearer resource_metadata="..."` headers from 401 responses for RFC 9728 §5.3 compliant discovery.
 - **Upstream introspection failures now return 401 instead of 503.** When Doorkeeper is unreachable, the previous behavior was `503 upstream_unavailable`. FastMCP's `IntrospectionTokenVerifier` treats transport failures as auth failures, so clients see 401. Operators monitoring 503 spikes as an upstream-Redmine-down signal should switch to monitoring 401-rate or watch `/health`'s new introspection probe (see Added).
 
 ### Added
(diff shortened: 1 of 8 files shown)


jztan:
@aadnehovda 
Thanks for the thorough work here, including the plugin and the end-to-end verification with Codex CLI and Claude Desktop. Before merging, I want to talk through the scope.
  
  The PR bundles two changes that have very different audiences:
  **1. Path rename** from /.well-known/oauth-authorization-server to /.well-known/oauth-authorization-server/mcp. This affects all OAuth-mode users today, is a clean hygiene improvement, and stands on its own.
  **2. Discovery mirroring with DCR pass-through.** This only takes effect when Redmine actually serves a discovery document, which today means installing redmine_oidc_provider.

  I checked the Redmine roadmap before responding. The relevant items I found are #43023 (IMAP via OAuth2, 7.0.0), #43352 (login to Redmine via external OAuth provider, 7.0.0), and #37363 (native OIDC, unassigned, open ~4 years). All three are about Redmine acting as an
  OAuth/OIDC client, not as a provider. There is no roadmap item, open ticket, or community discussion working toward Redmine serving RFC 8414 AS metadata or RFC 7591 DCR. So the discovery-mirroring code's only beneficiary today, and for the foreseeable future, is a Redmine
   instance running a third-party OIDC provider plugin.

  That brings me to a few concerns I'd like to work through with you:

  On the mirroring code path for stock Redmine

  For deployments without an OIDC provider plugin (which is ~all current deployments), every request to /.well-known/oauth-authorization-server/mcp triggers up to two HTTP calls to Redmine that will always fail, then falls through to the hardcoded defaults. Could we either
  cache the discovery result (5-15 min TTL) or short-circuit the fetch when an env var signals "stock Redmine, skip discovery"? That would remove the per-request cost for the majority case while keeping the DCR path for deployments that want it.

  **On redmine_oidc_provider**
  I want to understand your plans for the plugin. From what I can see, it's about a day old, no declared license and is self-described as experimental. A few questions:

  - Are you planning to maintain it long-term?
  - Will you add a license file? Without one, the default is "all rights reserved" and adopters can't legally use it.
  - Have you considered opening a Redmine issue or feature request to discuss adding OIDC-provider functionality to Redmine core or to a more established plugin like redmine_oauth?

  I'm not asking these to gatekeep, just to set expectations. If the MCP server ships plumbing for a plugin, the MCP server is implicitly betting on that plugin's trajectory.

  **Proposal**
  Would you be open to splitting this into two PRs?

  1. Path rename only, which I'd be happy to review and merge quickly. The CHANGELOG should call it out as a breaking change for any client hardcoding the old path, and the documentation updates you already have can come along with it.
  2. Discovery mirroring + DCR pass-through, which I'd want to hold until we've discussed the caching question, the plugin's maturity story, and (ideally) seen a license file land on redmine_oidc_provider.

  If splitting feels like extra work, an alternative is to keep the PR bundled but gate the mirroring behind an env var (default off), so stock-Redmine users pay no cost and deployments running an OIDC plugin opt in explicitly.

  Either way, happy to discuss. I appreciate the work, and I want to find a shape for this that ships value without coupling the project to maintenance surface that may not pay off.

jztan (changes requested):
See review comment above.

aadnehovda:
Thanks for the quick response!

I vibe coded up the plugin since it seems like a small natural addition to redmine. I have no idea whether anyone wants to use redmine as their idp (doubt it), so the need is perhaps not there, but the alternative to getting DCR to work was to use oauthproxy in fastmcp which seems more invasive, esp. when it comes to state management and that the mcp server now uses remoteoauth, not oauthproxy (which makes sense since redmine is the AS anyway and DCR probably belongs there). Ultimately it would be nice to also support CIMD to let online agents register statelessly, but I can find no mention about that in either doorkeeper or doorkeeper_openid_connect and afaict requires deeper changes in both due to the semantic changes to appid.

The oidc provider probably belongs in redmine core together with the other doorkeeper stuff, not in redmine_oauth which is redmine-as-idp-client -- the other direction. But again, maybe noone wants it (perhaps except for me to get DCR, it does make the registration very easy). I haven't thought about the license, but I'll just use the same license redmine uses. I'll maintain the plugin, answer issues and PRs against it, but I'm no oidc lowlevel expert.

When it comes to the autodiscovery I actually first considered gating the oidc support behind a static knob like an ENV var and not perform autodiscovery at all, but since the discovery should only happen during registration/login/token refresh and not for every tool call I thought it was cheap enough to keep it dynamic. Since it just hits the standard oidc endpoints, the mcp server should stay forward compliant should oidc provider support land in redmine core one day. But I agree the safest option is probably to opt-in to the oidc discovery altogether, since it depends on a clean failure mode for those oidc endpoints to actually fallback (which should ofc just return 404, but who knows for every setup).

I'm ofc happy to split it into two PRs.

aadnehovda:
Split as requested. This PR is now path rename only; the opt-in discovery mirroring follow-up is #136 and defaults off via REDMINE_MCP_MIRROR_REDMINE_AS_METADATA.

jztan (approved):
Clean split as discussed. Path move aligns with RFC 8414 §3 for a resource at /mcp, docs and tests updated symmetrically, all within the unreleased section. Approving and merging.

jztan:
Merged, thanks @aadnehovda for the quick split and the clean follow-up. Appreciate the collaboration on getting the scope right.

PR #136, opened 2026-05-23
Optionally mirror Redmine AS metadata

aadnehovda:
## Summary

Follow-up to #135. Until #135 merges, GitHub will show the path-scoped metadata commit in this PR too; the new change here is the second commit, `Optionally mirror Redmine AS metadata`.

This adds opt-in Redmine AS/OIDC metadata mirroring for deployments where Redmine itself serves discovery metadata:

- new env flag: `REDMINE_MCP_MIRROR_REDMINE_AS_METADATA=true`
- default remains static stock-Doorkeeper metadata, with no Redmine discovery fetches
- when enabled, the MCP server fetches Redmine discovery metadata from:
  - `/.well-known/oauth-authorization-server`
  - `/.well-known/openid-configuration`
- the mirrored metadata is then narrowed to the MCP resource's `scopes_supported`
- if Redmine discovery is unavailable, the server falls back to static Doorkeeper metadata without optional capabilities such as DCR

This keeps stock Redmine deployments on the zero-fetch path while allowing deployments with an OIDC-provider plugin or future Redmine-native provider support to advertise capabilities like `registration_endpoint`.

## Notes

One compatible Redmine plugin option is `redmine_oidc_provider`. That plugin is intentionally not required by this MCP server; this PR only mirrors standard discovery metadata when an operator explicitly opts in.

## Verification

```bash
UV_CACHE_DIR=/tmp/uv-cache-redmine-mcp uv run --extra dev pytest tests/test_oauth_auth.py tests/test_oauth_discovery.py tests/test_auth_factory.py
# 21 passed

UV_CACHE_DIR=/tmp/uv-cache-redmine-mcp uv run --extra dev black --check --fast src/redmine_mcp_server/main.py tests/test_oauth_discovery.py tests/test_oauth_auth.py
# 3 files would be left unchanged
```


jztan:
Hi @aadnehovda, first off thank you for the careful work here, and for being so willing to split #135 out and iterate. I really appreciate it, and #135 has already shipped value on its own.

I've been sitting with this one and I'd like to hold off on merging for now, not because of anything wrong with the code (it's clean, well-tested, and the default-off design directly addresses what I raised earlier), but because I don't yet have a user I can point at for the mirroring path. From our earlier discussion and your own honest read ("I have no idea whether anyone wants to use redmine as their idp"), the realistic beneficiary today is \`redmine_oidc_provider\`, and the broader Redmine ecosystem hasn't signaled movement toward provider-side OIDC.

My hesitation is about the long tail of maintenance rather than the change itself. Once a code path lives in the project, every future refactor of the discovery handler has to keep both branches working, and the docs that point at the plugin become something I'm implicitly endorsing over time. I'd rather wait until there's a concrete adopter (an issue filed, or \`redmine_oidc_provider\` picking up traction, or Redmine core moving on provider-side OIDC) and then merge this quickly when that signal arrives.

Concretely, I'd like to leave the branch open and the PR in a holding pattern rather than close it. If a user surfaces in the next few months, this becomes obvious to merge with minimal rework. If you'd prefer I close it for cleanliness and you reopen later, I'm happy to do that instead, your call.

Either way, please don't read this as a no to the work itself or to future contributions. The OAuth path in this project is meaningfully better because of your involvement, and I want to keep that going.

aadnehovda:
No worries :-), I don't think oidc in redmine is a good idea after all -- just to support DCR -- and since afaict CIMD in doorkeeper_openid-connect is not on the horizon, it would require yet another vanilla redmine plugin... The better option is to add OAuthProxy as a third auth mode here and accept some key persistence.

aadnehovda:
@jztan, FYI and if you are interested there's `OAuthProxy` auth-mode at aadnehovda/redmine-mcp-server@d0c09b3e0a461fbaafa763844bc345bfdd89afb0 .

jztan:
Thanks @aadnehovda, the work in `d0c09b3` is solid. Opt-in mode kept clean, and `require_authorization_consent="external"` is exactly the confused-deputy hardening I'd want before proxying a non-DCR upstream. Confirmed it works on the current FastMCP pin, no bump needed.

To be accurate on roadmap fit: `OAuthProxy` isn't in conflict with it. It's a FastMCP-native primitive, so it's within the direction v2.1 already committed to, and being opt-in it doesn't trigger the "avoid a second cutover" concern. It's just unlisted. It's also complementary rather than competing: the `RemoteAuthProvider` path has no DCR and Doorkeeper has no registration endpoint, so a DCR-only client can't get a `client_id` through discovery, in v2.1 or after the v3.0 hardening. `OAuthProxy` is the only thing that bridges that gap.

So I'd like to take it forward. Could you open a standalone PR for just the `oauth-proxy` mode (with its tests and the docs/env updates)? Keeping it separate from the #140 issuer fix, which is already in review as a minimal patch, lets the proxy mode go through review on its own merits, and I'm happy to help shape the env-var surface as part of that.

Thanks for pushing the OAuth story forward here.

aadnehovda:
Sounds good, I'll open the PR. It also includes some refactoring of the `RemoteAuthProvider` mode, specifically subclassing it as `RedmineAuthProvider` while folding in the AS and revoke endpoints. And support for mounting behind a prefix which should work for both oauth modes. The .well-known paths handling is somewhat messy upstream in both the SDK and FastMCP (as #140 highlights as well as the other recent changes) so this probably needs to be rebased and for sure tested well to make sure different clients are not broken.

jztan:
Hi @aadnehovda, no rush on this at all. Since we last spoke I cut v2.2.0, so `develop` has moved a fair bit (the #140 issuer fix, the #139 health/`current_user` work, and a few dependency bumps). I realize that makes the rebase for the proxy branch heavier than when you based it on `d0c09b3`. If it'd help, I'm happy to take the rebase onto current `develop` off your plate, or pair on the messy `.well-known` path handling you flagged, whatever makes it easiest to get the PR up. And once it's up I'm glad to run it through a multi-client end-to-end check: Claude Code on my side, plus a second client, so we can confirm the flow completes across more than one client.

PR #139, opened 2026-05-29
feat: expose current_user in get_mcp_server_info; probe Redmine in legacy /health

### LINKED ISSUE #138: get_mcp_server_info should expose current_user; /health should probe Redmine in legacy mode
## Problem

Two related usability gaps surface when an LLM uses the MCP server interactively:

### 1. `assigned_to_id=\"me\"` silently returns 0 results with a shared/robot API key

When `REDMINE_API_KEY` belongs to a robot or shared account (not the human operator), `list_redmine_issues(assigned_to_id=\"me\")` returns 0 results with no diagnostic information. The LLM has no way to know that \"me\" resolves to the robot account rather than the human user — it just sees an empty list and may incorrectly conclude there are no open issues.

### 2. `/health` does not verify Redmine connectivity in legacy mode

In legacy mode, `GET /health` always returns `{\"status\": \"ok\"}` regardless of whether the configured API key actually works. Auth failures are invisible until a tool is called, which makes diagnosing misconfigured deployments much harder.

## Proposed fix

**`tools/meta.py` — add `current_user` to `get_mcp_server_info`:**

Add `current_user: {id, login, name}` to the response of `get_mcp_server_info`. An LLM can call this tool first to immediately see who `\"me\"` maps to, instead of getting a silent empty result.

**`tools/issues.py` — clarify `assigned_to_id` docstring:**

Note that `\"me\"` resolves to the API key owner, not necessarily the human operator. Direct callers to `get_mcp_server_info` when results are unexpectedly empty.

**`_http_routes.py` — probe Redmine in legacy mode health check:**

In legacy mode, `GET /health` now calls `GET /my/account.json` to verify the API key. If credentials are configured but the call fails → `status: degraded`. If no credentials are configured → `status: ok` with `checks.redmine: \"unconfigured\"` (not yet configured, not a runtime failure).

## Example

Before:
```json
// GET /health (legacy mode)
{"status": "ok", "service": "redmine_mcp_tools", "auth_mode": "legacy"}

// get_mcp_server_info
{"server_version": "...", "read_only_mode": false, "auth_mode": "legacy", "plugin_flags": {...}}
```

After:
```json
// GET /health (legacy mode, credentials configured and working)
{"status": "ok", "service": "redmine_mcp_tools", "auth_mode": "legacy",
 "checks": {"redmine": "ok"}}

// GET /health (legacy mode, auth failure)
{"status": "degraded", "service": "redmine_mcp_tools", "auth_mode": "legacy",
 "checks": {"redmine": "unreachable", "redmine_detail": "AuthError"}}

// get_mcp_server_info
{"server_version": "...", "read_only_mode": false, "auth_mode": "legacy",
 "current_user": {"id": 5, "login": "vitex", "name": "Vítězslav Dvořák"},
 "plugin_flags": {...}}
```

A PR with tests is ready.

Vitexus:
Closes #138

## Summary

- **`get_mcp_server_info`** now includes `current_user: {id, login, name}` so an LLM can immediately see who `assigned_to_id="me"` maps to — crucial when a shared or robot API key is in use.
- **`list_redmine_issues` docstring** clarifies that `"me"` resolves to the API key owner, not necessarily the human operator, and directs callers to `get_mcp_server_info` when results are unexpectedly empty.
- **`GET /health` (legacy mode)** now calls `GET /my/account.json` to verify the API key works:
  - credentials configured and Redmine reachable → `checks.redmine: "ok"`
  - credentials configured but Redmine unreachable/auth failure → `status: "degraded"`
  - no credentials configured → `checks.redmine: "unconfigured"`, `status: "ok"` (not yet set up, not a runtime failure)

## Test plan

- [ ] `test_server_info.py` — 4 new assertions: `current_user` is `None` when unreachable; `{id, login, name}` when connected; key set pinned to include `current_user`; credential leak check still passes
- [ ] `test_health_probe.py` — renamed `test_legacy_mode_skips_introspection_probe` → `test_legacy_mode_probes_redmine_not_introspection`; added `test_legacy_mode_degraded_when_redmine_unreachable`
- [ ] Full unit suite: `PYTHONPATH=src pytest tests/ --ignore=tests/test_ssl_integration.py --ignore=tests/test_integration.py --ignore=tests/test_oauth_integration.py` → 1289 passed, 4 skipped

### DIFF
--- a/src/redmine_mcp_server/_http_routes.py
+++ b/src/redmine_mcp_server/_http_routes.py
@@ -93,7 +93,7 @@ async def _probe_introspection() -> tuple[str, Optional[str]]:
 async def _probe_redmine_legacy() -> tuple[str, str | None]:
     """Check Redmine connectivity for legacy (API key / password) auth mode.
 
-    Calls ``GET /my/account.json`` using the configured credentials.
+    Calls ``GET /users/current.json`` using the configured credentials.
 
     Returns:
         ``("ok", None)`` — credentials valid and Redmine reachable.
@@ -150,7 +150,7 @@ async def health_check(request):
     upstream availability that was lost in the 503->401 collapse when
     FastMCP native auth replaced the bespoke middleware.
 
-    In legacy mode, probes ``GET /my/account.json`` to verify the configured
+    In legacy mode, probes ``GET /users/current.json`` to verify the configured
     API key (or username/password) is accepted by Redmine.
 
     Returns HTTP 200 in both healthy and degraded states so container
--- a/src/redmine_mcp_server/_http_routes.py
+++ b/src/redmine_mcp_server/_http_routes.py
@@ -116,10 +116,12 @@ async def _probe_redmine_legacy() -> tuple[str, str | None]:
     if not (REDMINE_API_KEY or (REDMINE_USERNAME and REDMINE_PASSWORD)):
         return "unconfigured", "no credentials configured"
 
-    # Use httpx directly against /my/account.json.
-    # redminelib's user.get('current') hits /users/current.json which requires
-    # admin rights and fails for non-admin API keys.
-    url = REDMINE_URL.rstrip("/") + "/my/account.json"
+    # Use httpx directly against /users/current.json.
+    # This endpoint works on Redmine 3.x and later (/my/account.json is
+    # not reliably available on older instances).
+    # redminelib's user.get('current') is NOT used here because it requires
+    # admin rights on some Redmine setups.
+    url = REDMINE_URL.rstrip("/") + "/users/current.json"
     try:
         if REDMINE_API_KEY:
             headers = {"X-Redmine-API-Key": REDMINE_API_KEY}
--- a/src/redmine_mcp_server/_http_routes.py
+++ b/src/redmine_mcp_server/_http_routes.py
@@ -109,7 +109,6 @@ async def _probe_redmine_legacy() -> tuple[str, str | None]:
         REDMINE_PASSWORD,
         REDMINE_URL,
         REDMINE_USERNAME,
-        _get_redmine_client,
     )
 
     if not REDMINE_URL:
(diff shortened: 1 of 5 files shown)


jztan (changes requested):
Thanks for this, @Vitexus. Both gaps in #138 are real and the tests are thorough. The branch is cleanly scoped. Two changes before merge.

**1. `get_mcp_server_info` does blocking I/O on the event loop.**

`get_mcp_server_info` is an async tool, but `_fetch_current_user_info()` calls `requests.get(..., timeout=5)` synchronously, so a slow or unreachable Redmine blocks the event loop for up to 5 seconds. Your `_probe_redmine_legacy` in `_http_routes.py` already does this correctly with async `httpx`. Switching `meta.py` to async `httpx` fixes two things at once:

- no more event-loop blocking, and
- `requests` goes away. It isn't a declared dependency (`httpx` is, and it's what the rest of the codebase uses), so dropping `import requests` removes that transitive reliance.

Since both call `/users/current.json` with identical auth logic, they could share a single async helper.

**2. Stale docstring.** `_probe_redmine_legacy`'s docstring still says *"Calls `GET /my/account.json`"*, but the code now calls `/users/current.json`. Please sync it so it doesn't mislead later.

Once both are in, I'm happy to merge. Thanks again.

Vitexus:
Thanks for the thorough review, @jztan. Both points addressed in b322c49:

**1. `_fetch_current_user_info` → async httpx**
Dropped `import requests` entirely and switched to `async with httpx.AsyncClient(timeout=5)` — same pattern as `_probe_redmine_legacy`. The function is now `async` and the call site uses `await`.

**2. Stale docstrings synced**
Fixed both places that still said `GET /my/account.json`:
- `_probe_redmine_legacy` docstring
- `health_check` docstring

**Live test against both target Redmine versions**

Ran `_fetch_current_user_info` and `_probe_redmine_legacy` directly from source against the two production instances:

| Instance | Redmine era | `current_user` result | health probe |
|---|---|---|---|
| crm.spoje.net | 3.x (Passenger 5.0.30 / Apache 2.4.25) | `{id: 43, login: "vitex", name: "Vítězslav Dvořák"}` | `ok` |
| redmine.vitexsoftware.com | 6.x (Apache 2.4.67) | `{id: 14, login: "api", name: "Robot "}` | `ok` |

Both the async httpx path and the probe return clean results on Redmine 3.x and 6.x.

Full unit suite still at 1289 passed, 4 skipped.

jztan (approved):
Both review points are fully addressed: `_fetch_current_user_info` is now async `httpx` (no more event-loop blocking, `requests` dropped), and the `/users/current.json` docstrings are consistent. Ran the suite locally: `test_server_info.py` and `test_health_probe.py` pass (14/14), flake8 clean. Thanks for the quick turnaround.

jztan:
Merged, thanks @Vitexus! This is a genuinely useful pair of fixes: surfacing `current_user` makes the shared/robot-key "me" ambiguity obvious to an LLM caller, and the legacy `/health` probe means auth misconfig shows up at the health check instead of on the first failed tool call. Appreciate the quick turnaround on moving to async `httpx` and the test coverage.

PR #177, opened 2026-07-08
feat(issues): expand full-issue serializer with standard fields (#174)

### LINKED ISSUE #174: get_redmine_issue: include standard issue fields (fixed_version, category, dates, …) in the full-issue serializer
### Problem Statement

`get_redmine_issue` is documented as pulling "a single issue with **full context**", but its serializer (`_issue_to_dict`) returns only a curated subset of fields. Several fields that Redmine's default issue JSON already returns are dropped, so an MCP client cannot see them:
  
  - `fixed_version` (target version)
  - `category`
  - `start_date`, `due_date`, `closed_on`
  - `done_ratio`
  - `estimated_hours`, `spent_hours`
  - `parent`
  - `is_private`

These are core Redmine fields (no plugin dependency) and are routinely needed to reason about an issue — which release it targets, which category/component, its scheduling/progress, and its parent/subtask structure. Today the  only way to get them is to bypass the MCP and hit Redmine's REST API directly.
  
There is also an internal inconsistency: the sibling `_gantt_issue_to_dict` already serializes several of these (`start_date`, `due_date`, `estimated_hours`, `parent`) — its own comment notes it includes fields "that `_issue_to_dict` omits". So the dedicated full-issue tool returns *less* than the gantt tool.

### Proposed Solution

  Add the missing standard fields to `_issue_to_dict` (and mirror them in `_issue_to_dict_selective`'s field map so field selection can request them):

  - `fixed_version` → `{id, name}`
  - `category` → `{id, name}`
  - `parent` → `{id}`
  - `start_date`, `due_date`, `closed_on` (ISO strings)
  - `done_ratio`, `estimated_hours`, `spent_hours` (numbers)
  - `is_private` (bool)
  - 
All of these are already present on the python-redmine issue object returned by the default `issue.get`, so no extra HTTP request is needed. Structured references follow the existing `{id, name}` convention used for `project`/`status`/etc.; none are free-text, so no `<insecure-content>` wrapping is required.

### Use Case

An agent triaging a backlog asks: "Which release is issue X targeted at, and is it scheduled?" With the current output it cannot tell — `fixed_version`, `category`, and `start_date`/`due_date` are all absent — so it either answers incorrectly or has to leave the MCP and query Redmine's REST API directly. With these fields included, a single `get_redmine_issue` call is enough to reason about targeting, categorization, scheduling, progress, and parent/subtask context.

### Feature Category

Existing Tool Enhancement

### Priority

Important

### Alternatives Considered

 - Falling back to the raw Redmine REST API (`/issues/<id>.json`) for these fields — works, but defeats the purpose of the MCP and needs separate credentials/handling.
  - Using `get_gantt_chart` for scheduling fields — awkward and only partial (no `fixed_version` / `category`).

### Additional Context

All listed fields are confirmed present on the default `/issues/<id>.json` response and exposed by python-redmine via `getattr`. 

### Contribution

LaurensRietveld:
## Summary

Fixes #174.

`get_redmine_issue` advertises "full context", but its `_issue_to_dict` serializer dropped several core fields that Redmine's default issue JSON already returns — forcing clients to fall back to the raw REST API for triage and scheduling. The sibling `_gantt_issue_to_dict` serializer already exposed a subset (dates, estimated hours, parent), so the dedicated full-issue tool ironically returned *fewer* fields than the gantt tool.

This adds the missing standard fields to `_issue_to_dict`:

- `category` — `{id, name}`
- `fixed_version` (target version) — `{id, name}`
- `parent` — `{id}` (matches Redmine's raw JSON shape)
- `start_date`, `due_date`, `closed_on` — ISO-8601 via the existing `_safe_isoformat`
- `done_ratio`, `estimated_hours`, `spent_hours`
- `is_private`

All values are already present on the python-redmine object, so **no additional HTTP request** is made. Reference fields reuse the existing `{id, name}` convention, and every field degrades to `None` when absent (search API may return sparse objects).

The same fields are mirrored in `_issue_to_dict_selective` (and its docstring) so `list_redmine_issues` / `search_redmine_issues` field selection stays consistent and these fields are individually selectable.

## Testing

- Added tests for populated values, the `None`-when-absent path, and individual field selection.
- Updated existing "all fields" count assertions (11 → 21).
- `uv run pytest -m "not integration" tests/` → **1404 passed**.
- `black --check src/` and `flake8 src/` clean.

### DIFF
--- a/src/redmine_mcp_server/tools/issues.py
+++ b/src/redmine_mcp_server/tools/issues.py
@@ -238,12 +238,9 @@ def _issue_to_dict(issue: Any, include_custom_fields: bool = False) -> Dict[str,
             if assigned is not None
             else None
         ),
-        # Standard fields returned by Redmine's default issue JSON. These carry
-        # workflow state clients routinely need (release targeting via
-        # fixed_version, categorization, scheduling dates, progress and effort)
-        # and are already present on the python-redmine object, so no extra
-        # HTTP request is needed. The sibling gantt serializer already exposed
-        # a subset of these; see GitHub issue #174.
+        # Standard fields returned by Redmine's default issue JSON.
+        # The sibling gantt serializer already exposes a subset of these.
+        # see GitHub issue #174.
         "category": (
             {"id": category.id, "name": category.name} if category is not None else None
         ),
--- a/src/redmine_mcp_server/tools/issues.py
+++ b/src/redmine_mcp_server/tools/issues.py
@@ -207,6 +207,9 @@ def _issue_to_dict(issue: Any, include_custom_fields: bool = False) -> Dict[str,
     priority = getattr(issue, "priority", None)
     author = getattr(issue, "author", None)
     tracker = getattr(issue, "tracker", None)
+    category = getattr(issue, "category", None)
+    fixed_version = getattr(issue, "fixed_version", None)
+    parent = getattr(issue, "parent", None)
 
     issue_dict = {
         "id": getattr(issue, "id", None),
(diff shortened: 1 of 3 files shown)


jztan (approved):
Thanks @LaurensRietveld, this is a clean fix and a good catch that the "full context" tool was returning fewer fields than the gantt serializer. The implementation follows the existing conventions nicely: the `{id, name}` reuse, the `None`-when-absent guards, and reading straight off the already-loaded issue object so there's no extra Redmine round-trip. Test coverage is thorough too, covering populated, absent, and individual field selection. Merging this. I'll follow up on the docs and changelog separately.

PR #178, opened 2026-07-08
feat: optional additional_tags plugin support (read + write)

### LINKED ISSUE #175: Optional support for the additional_tags plugin
### Problem Statement

Many Redmine deployments use the AlphaNodes `additional_tags` plugin and encode workflow state as issue tags (e.g. a `fast-track` tag). The plugin adds a `tags` array to the default issue JSON, but that state is invisible through the MCP
  
This is analogous to the existing agile support: agile is a plugin whose fields + scope are opt-in via `REDMINE_AGILE_ENABLED`. There's no equivalent for tags. 

### Proposed Solution

Add opt-in support for `additional_tags`, gated behind a new env flag (e.g. `REDMINE_TAGS_ENABLED`), mirroring `REDMINE_AGILE_ENABLED`

### Use Case

Our team encodes workflow state as issue tags (e.g. `fast-track`). An agent asked "is issue X fast-tracked?" can't tell, because `tags` never reaches the client. With the flag on and `view_issue_tags` in the token, `get_redmine_issue` returns `tags: [{"name": "fast-track"}, ...]`.

### Feature Category

New MCP Tool

### Priority

Nice to have

### Alternatives Considered

Using the API directly

### Additional Context

Can imagine you won't want the support all possible plugins in this MCP. Hence, the 'nice to have'.

Some other thoughts: for plugins that ride on the issue json, you could consider supporting more generic options like:
- `REDMINE_EXTRA_SCOPES` for influencing the scopes the MCP requests
- Passing-through of unknown issue-json scalar fields (possibly as opt-in feature). 

But guess these issues are all up to you on how you want to scope this MCP going forward

### Contribution

LaurensRietveld:
Closes #175.

Opt-in support for the AlphaNodes [`additional_tags`](https://github.com/alphanodes/additional_tags) plugin, mirroring the existing `REDMINE_AGILE_ENABLED` pattern. Off by default; everything below is gated on `REDMINE_TAGS_ENABLED`.

## What it does

**Read** — `get_redmine_issue` returns a `tags` array when enabled:
```json
"tags": [{"id": 3, "name": "fast-track"}, {"id": null, "name": "backend"}]
```
The plugin injects `tags` directly into `GET /issues/{id}.json` (only on `show`, and only when the caller holds `view_issue_tags`), so no extra request is needed — the field is read off the issue object already fetched. The plugin often omits the tag `id` (it only emits it when the sorted `tag_list` matches its tag records), so `id` is normalized to `null` when absent; `name` is the stable identifier. Empty list when the issue has no tags or the caller lacks the permission.

**Write** — `create_redmine_issue` / `update_redmine_issue` accept a `tag_list` (a list of names, or a comma-separated string; `[]` clears). It's extracted *before* custom-field name resolution so it can't collide with a same-named custom field, and it replaces the issue's full tag set. Silently ignored when the flag is off (same as `story_points` without the agile flag).

**OAuth scopes** — advertised only when enabled, so a Redmine without the plugin never sees an unresolvable scope:
- `view_issue_tags` (read) — kept in read-only mode.
- `create_issue_tags` / `edit_issue_tags` (write) — dropped in read-only mode.

**Discovery** — `get_mcp_server_info` reports `plugin_flags.tags`.

## Testing

- Unit tests: env flag, tag serialization (id present/absent/object/empty), create/update `tag_list` handling + disabled-drop, comma-string normalization, and all four scope-advertisement cases (`tests/test_tags_support.py`, `tests/test_oauth_scopes.py`, `tests/test_server_info.py`). Full suite green.
- Integration tests (opt-in, gated on `REDMINE_TAGS_ENABLED` / a sandbox project): read shape + a create→update→read→delete `tag_list` round-trip (`tests/test_integration.py`).
- Verified end-to-end against a live Redmine running `additional_tags` v4.4.0: create with a comma-string `tag_list` applied both tags, `tag_list: []` cleared them, the change was recorded in the issue journal, and read-back returned the `tags` array.

## Docs

README env table, `docs/tool-reference.md` (get/create/update + `get_mcp_server_info`), `docs/troubleshooting.md` (missing-tags and silent-write sections), `docs/contributing.md`, and both `.env` examples.


jztan:
Thanks @LaurensRietveld, merged. This is a clean, well-tested implementation that mirrors the existing agile-plugin pattern nicely, and I appreciated the end-to-end verification against `additional_tags` v4.4.0.

I pushed one small follow-up before merging: `tag_list` is now re-applied on the required-custom-field autofill retry path (when `REDMINE_AUTOFILL_REQUIRED_CUSTOM_FIELDS=true`). It was being silently dropped on that retry, the same way agile `story_points` is re-applied, plus a regression test for both create and update. Changelog credit is in under `[Unreleased]`.

PR #179, opened 2026-07-13
Dockerfile: respect SERVER_HOST and SERVER_PORT env vars

pdostal:
The Dockerfile CMD hardcoded `--host 0.0.0.0 --port 8000` directly in the uvicorn invocation, bypassing `main()` and making `SERVER_HOST`/`SERVER_PORT` env vars completely ineffective when running the container.

Switch CMD to the installed `redmine-mcp-server` console script, which already reads both vars from the environment. Also make the HEALTHCHECK port dynamic via `${SERVER_PORT:-8000}` and update the EXPOSE comment to clarify it's a default hint only.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,10 +59,10 @@ USER appuser
 
 # Health check
 HEALTHCHECK --interval=30s --timeout=10s --start-period=5s --retries=3 \
-    CMD curl -f http://localhost:8000/health || exit 1
+    CMD curl -f http://localhost:${SERVER_PORT:-8000}/health || exit 1
 
-# Expose port
+# Expose default port (informational only; override with SERVER_PORT env var)
 EXPOSE 8000
 
 # Default command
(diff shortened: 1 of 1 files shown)


jztan (approved):
Thanks @pdostal, nice catch. The hardcoded `--host 0.0.0.0 --port 8000` really did make `SERVER_HOST`/`SERVER_PORT` dead in the container, and moving to the `redmine-mcp-server` console script is the right fix. I confirmed the script resolves from `/opt/venv` on `PATH` and that `main()` reads both vars.

One thing I will follow up on myself: `main()` defaults `SERVER_HOST` to `127.0.0.1`, whereas the old `CMD` hardcoded `0.0.0.0`, so an ad-hoc `docker run` without a full env file would now bind to loopback. I will add an `ENV SERVER_HOST=0.0.0.0` default in a follow-up so the out-of-the-box image stays reachable. That does not block this change since the documented compose flow sets it via `.env.docker`.

Merging now. Thanks for the fix.

jztan:
Merged to `develop`. Thanks again @pdostal for fixing the container env var handling.

PR #194, opened 2026-07-27
feat: support writing agile_sprint_id/position via update_redmine_issue

### LINKED ISSUE #193: update_redmine_issue ignores agile_sprint_id
### Bug Description

`update_redmine_issue` silently ignores `agile_sprint_id`, so an issue cannot be moved to
or removed from a sprint through the MCP server. With `REDMINE_AGILE_ENABLED=true`, the
function extracts and writes `story_points` to the RedmineUP Agile plugin, but no other
agile field is handled. Passing `agile_sprint_id` — either top-level or nested as
`fields={"agile_data_attributes": {"agile_sprint_id": <id>}}` — is dropped: the call
returns a success response while the sprint is unchanged, so callers get a false positive.
(`_fetch_agile_data` already *reads* `agile_sprint_id`, so the gap is only on the write path.)

### Steps to Reproduce

1. Configure the server with `REDMINE_AGILE_ENABLED=true` against a Redmine with the
   RedmineUP Agile plugin.
2. Call `update_redmine_issue(issue_id=<id>, fields={"agile_data_attributes": {"agile_sprint_id": <sprint_id>}})`
   (also fails with a top-level `agile_sprint_id`).
3. The response indicates success, but the issue's sprint is unchanged. `get_redmine_issue`
   still reports the old `agile_sprint_id`.

### Expected Behavior

Setting `agile_sprint_id` through `update_redmine_issue` moves the issue to that sprint, and
`agile_sprint_id: 0`/null removes it from the sprint — the same way `story_points` is written.
Write failures should surface an error instead of a silent success.

### Actual Behavior

`agile_sprint_id` is passed to python-redmine's core issue update (or treated as a candidate
custom field), which does not understand `agile_data_attributes`, so the RedmineUP Agile
plugin never receives it. The tool reports success and the sprint is unchanged.

### Version

2.8.0

### Deployment Method

PyPI installation (uv/pip)

### MCP Client

Claude Desktop

### Redmine Version

6.1.13

### Relevant Logs

```shell
no error is produced; the call returns a normal success payload
```

### Additional Context

A direct `PUT /issues/{id}.json` with
`{"issue": {"agile_data_attributes": {"agile_sprint_id": <id>}}}` sets/clears the sprint
correctly — the same endpoint the existing `story_points` write already uses. A PR
generalizing the agile-write path is ready.

knasiotis:
## Description
`update_redmine_issue` could write `story_points` to the RedmineUP Agile plugin but silently
dropped `agile_sprint_id`, so issues could not be moved to/from a sprint through the MCP server
(the call returned success while the sprint was unchanged). This generalizes the agile-write
path so sprint and position are writable too, and echoes the resulting agile fields back in the
response so the change is verifiable.

## Related Issue
Fixes #193

## Changes Made
- Generalize `_apply_agile_story_points` into `_apply_agile_data(issue_id, agile_attrs)`, which
  PUTs `{"issue": {"agile_data_attributes": <attrs>}}` to `/issues/{id}.json`.
  `_apply_agile_story_points` is kept as a thin back-compat wrapper.
- In `update_redmine_issue`, a nested `agile_data_attributes` dict in `fields` may now carry
  `agile_sprint_id` (`0`/null removes the issue from its sprint), `position`, and `story_points`.
  Top-level `story_points` still works unchanged. Agile keys are collected with explicit
  key-presence checks (so `0`/null still triggers the write) and popped before custom-field
  mapping so `agile_data_attributes` is never mistaken for a custom field. When
  `REDMINE_AGILE_ENABLED=false`, these keys are silently ignored as before.
- The updated issue is augmented with the resulting `story_points`, `agile_sprint_id`, and
  `agile_position` (mirroring `get_redmine_issue`), only when an agile field was actually
  written, so plain updates make no extra HTTP call.
- `get_redmine_issue`'s agile augmentation refactored to reuse the new `_augment_with_agile_data`
  helper.

## Testing
- [x] Unit tests added/updated (`tests/test_agile_support.py`: set sprint, remove-with-0,
  combined standard+sprint update, drop-when-disabled, response echo, `_apply_agile_data`)
- [x] Full non-integration suite passes locally: **1527 passed**
- [x] Manual: raw `PUT /issues/{id}.json` with `agile_data_attributes.agile_sprint_id`
  confirmed to set/clear the sprint (same endpoint the fix uses)
- [x] `black --check` and `flake8 --max-line-length=88` clean on changed files

## Checklist
- [x] Code follows style guidelines (Black, flake8 @ 88)
- [x] Documentation updated (docstring + CHANGELOG)
- [x] Tests added/updated
- [x] CHANGELOG updated

## Compatibility
Backwards compatible — top-level `story_points` and all existing behavior are unchanged; only
the previously-dropped agile keys gain effect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,15 +9,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - Write support for agile sprint and position via `update_redmine_issue`
-  ([#193](https://github.com/jztan/redmine-mcp-server/issues/193)): a nested
-  `agile_data_attributes` dict in `fields` may now carry `agile_sprint_id`
-  (set to `0`/null to remove the issue from its sprint), `position`, and
-  `story_points`, routed to the RedmineUP Agile plugin endpoint the same way
-  top-level `story_points` already was. Previously `agile_sprint_id` was
+  ([#193](https://github.com/jztan/redmine-mcp-server/issues/193)):
+  `agile_sprint_id` (set to `0`/null to remove the issue from its sprint) and
+  `position` may now be set, given either top-level in `fields` or nested under
+  an `agile_data_attributes` dict, routed to the RedmineUP Agile plugin endpoint
+  the same way `story_points` already was. Previously `agile_sprint_id` was
   silently dropped, so issues could not be moved to/from a sprint through the
-  MCP server. The updated issue is now augmented with the resulting
-  `story_points`, `agile_sprint_id`, and `agile_position` so the change can be
-  verified from the response.
+  MCP server. Writes are applied in place — the current `agile_data` row (id and
+  existing values) is carried forward — so changing one agile field never nulls
+  the others. The updated issue is augmented with the resulting `story_points`,
+  `agile_sprint_id`, and `agile_position` so the change can be verified from the
+  response.
 
 ## [2.8.0] - 2026-07-25
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Write support for agile sprint and position via `update_redmine_issue`
+  ([#NNN](https://github.com/jztan/redmine-mcp-server/issues/NNN)): a nested
+  `agile_data_attributes` dict in `fields` may now carry `agile_sprint_id`
+  (set to `0`/null to remove the issue from its sprint), `position`, and
+  `story_points`, routed to the RedmineUP Agile plugin endpoint the same way
+  top-level `story_points` already was. Previously `agile_sprint_id` was
+  silently dropped, so issues could not be moved to/from a sprint through the
+  MCP server. The updated issue is now augmented with the resulting
+  `story_points`, `agile_sprint_id`, and `agile_position` so the change can be
+  verified from the response.
 
 ## [2.8.0] - 2026-07-25
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - Write support for agile sprint and position via `update_redmine_issue`
-  ([#NNN](https://github.com/jztan/redmine-mcp-server/issues/NNN)): a nested
+  ([#193](https://github.com/jztan/redmine-mcp-server/issues/193)): a nested
   `agile_data_attributes` dict in `fields` may now carry `agile_sprint_id`
   (set to `0`/null to remove the issue from its sprint), `position`, and
   `story_points`, routed to the RedmineUP Agile plugin endpoint the same way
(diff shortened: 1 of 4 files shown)


jztan (changes requested):
Thanks @knasiotis, the approach is right and #193 was a great report. Suite passes for me too (1527), black and flake8 clean.

I ran the branch against a live Redmine 6.1.1 and hit a blocker. The plugin's `accepts_nested_attributes_for :agile_data` has no `update_only: true`, so a payload without an `id` replaces the row instead of updating it, nulling every field you did not send:

```
seed  story_points=5, position=7   -> {story_points: 5, position: 7}   (row 29)
PUT   {story_points: 6}            -> {story_points: 6, position: None} (row 30)
PUT   {"id": 33, "story_points": 9} -> {story_points: 9, position: 7}   (row 33 kept)
```

So a sprint move via `{"agile_data_attributes": {"agile_sprint_id": N}}` wipes story points every time. `_apply_agile_data` needs to GET `agile_data.json` first and pass the id through, omitting it when the issue has no row yet (most report `id: null`). This is pre-existing for `story_points`, I will file it separately.

Two silent successes also remain, same class as #193: a top-level `agile_sprint_id` still goes nowhere and returns success, and an unrecognized key inside the nested dict is dropped with no write and no error.

Minor: `docs/tool-reference.md` still lists `story_points` as the only writable agile field, and `_apply_agile_story_points` is now only used by tests.

One thing I cannot check: my sandbox runs Agile Light, no sprints. Could you confirm on Pro that `issue_patch.rb` is the same there, and that `agile_sprint_id: 0` really clears membership? Light happily accepted a nonexistent sprint id.

Happy to take the id fix myself if you would rather not carry it.

knasiotis:
Hi thanks for the fast reply. Here is a summary of the latest changes.

**1. `accepts_nested_attributes_for :agile_data` without `update_only: true` (data loss).**
A nested payload without the row `id` builds a new `agile_data` record and
nulls the fields I didn't send. `_apply_agile_data` now reads the current row first and
carries its `id` and existing `story_points`/`agile_sprint_id`/`position` forward, then
overlays the requested change, so the write updates in place and setting one field no
longer wipes the others. When there's no existing row it falls back to a plain create.
Regression test: `test_updates_in_place_preserving_other_fields` (and the sprint-set/update
tests now assert the untouched fields survive in the PUT payload).

**2. Top-level `agile_sprint_id` went unprocessed.**
Fixed. `agile_sprint_id`, `position`, and `story_points` are now each accepted either
top-level in `fields` or nested under `agile_data_attributes` (equivalent), matching the
existing top-level `story_points`. Covered by `test_sets_sprint_via_top_level_key`.

**3. Documentation.**
`docs/tool-reference.md` now documents the writable agile fields, both spellings, the
in-place behavior, and the response echo; the tool docstring and CHANGELOG are updated too.

**4. Does `agile_sprint_id: 0` clear membership on Agile Pro?**
I can't verify this against Agile Pro — I don't have a Pro instance to test on. The tool
forwards whatever the caller sends, and the extraction uses an explicit key-presence check,
so both `0` and `null` reach the plugin (each triggers the write).

jztan (approved):
Thanks @knasiotis, that was quick. I re-checked against a live Redmine 6.1.1 rather than just reading the diff:

- In-place update confirmed. Seeded `story_points=5, position=7`, wrote `agile_sprint_id=3`, all three survived and the `agile_data` row id was reused.
- Top-level `agile_sprint_id` writes now, and no longer leaks into the core issue update.
- Thanks for documenting the custom-field collision on `position` and `agile_sprint_id`. I had not asked for that, and it matters, since `position` is a generic name that is now intercepted whenever the flag is on.

Suite and flake8 clean for me too. (Three unrelated test files already fail `black --check` on develop, nothing to do with this.)

Two small things I will pick up myself in a follow-up rather than hold this up:

- An unrecognized key inside `agile_data_attributes`, or a stringified one, is still dropped with no write and no error.
- The read-failure fallback in `_apply_agile_data` sets `current = {}`, which is exactly the id-less payload this commit exists to avoid, so a transient GET failure reverts to replacing the row. Narrowing the catch to not-found would keep the create path working without that.

Fair enough on Agile Pro, I cannot test that either.

jztan:
Merged, thanks @knasiotis. Tracking down the `accepts_nested_attributes_for` replacement behavior was the valuable part here, it was a live bug in 2.8.0 independent of the sprint feature.

PR #200, opened 2026-08-05
fix(wiki): serialize the dict-shaped project ref added in Redmine 7.0

azelcs:
## Summary

`manage_redmine_wiki_page` crashes against Redmine 7.0 with `An unexpected error occurred while fetching wiki page '<title>' in project <project>: 'dict' object has no attribute 'id'`.

Redmine 7.0 added a `project` ref to the wiki page API response ([Redmine #43569](https://www.redmine.org/issues/43569), one added line in `app/views/wiki/show.api.rsb`). python-redmine only wraps a nested key as a `Resource` when the parent lists it in its `_resource_map`, and `WikiPage._resource_map` is `{'author': 'User'}` — no `project`. So `project` arrives as a plain `dict`, and `_wiki_page_to_dict`'s `wiki_page.project.id` raises. The `AttributeError` is caught by the same handler that wraps the network call, which is why a serialization bug reads as a Redmine API failure.

This affects `get`, `create`, `update`, and `rename` (all serialize a page); `list` and `delete` are unaffected. The write actions are the nastier case — they applied the change and *then* failed, reporting an error for a write that had already succeeded. Redmine 6.1 and earlier are unaffected: they don't send the field.

Not fixable by a dependency bump — python-redmine 2.5.0 is the latest release and the pin is already `>=2.5.0,<3`.

Existing tests missed this because the fixtures build `project` as a `Mock()` with `.id`/`.name`, which behaves like a `Resource`.

## Changes

- `_named_ref` (`_serialization.py`) reads dict-shaped refs via `.get()` instead of `getattr`. It previously returned `{"id": None, "name": ""}` for a dict — silent data loss — so this also repairs six other call sites that pass potentially-dict refs (`tools/gantt.py`, `tools/files.py`, `tools/time_tracking.py`, `tools/projects.py`, `tools/issues.py`, and `_attachment_to_dict`).
- `_wiki_page_to_dict` (`tools/wiki.py`) routes both `author` and `project` through `_named_ref` instead of hand-rolled ref dicts. The `hasattr` guards are kept — they're what omits the keys entirely on pre-7.0 servers.
- Unit tests: new `tests/test_named_ref_dict_shape.py` for the helper; a dict-`project` regression test beside the existing wiki `get` tests in `tests/test_global_search.py`.
- Integration test: the wiki lifecycle test asserts that when `project` is present it carries a real id/name, and now also exercises `rename` — the fourth affected action, which previously had no integration coverage at all. The assertion is version-agnostic, so it passes on both 6.1 (key absent) and 7.0 (key present).
- `CHANGELOG.md` and `docs/tool-reference.md` (the `project` field is documented as Redmine 7.0+; it was previously listed unconditionally).

## Testing

The integration tests skip entirely unless `REDMINE_URL` is set, so reproducing this needs a `.env` pointing at a live Redmine. Run against containerized **7.0.0** (where the bug appears) and **6.1.2** (where the field is absent), identical results on both:

```
$ python tests/run_tests.py --all
  unit:        1539 passed, 87 deselected
  integration: 76 passed, 11 skipped   # skips = optional plugins + live OAuth

$ uv run black --check src/ --line-length=88     # clean
$ uv run flake8 src/ --max-line-length=88        # clean
```

The new unit tests and the wiki lifecycle integration test both fail before the fix and pass after. Live, all four affected actions (`create`, `get`, `update`, `rename`) return `project: {id, name}` on 7.0.0 and omit the key on 6.1.2 — the version-agnostic assertion holds in both directions.

Note: run `tests/fixtures/ssl/generate-test-certs.sh` first — the certs aren't in the repo (removed in 1d17604, generated in CI), so `test_ssl_cert_file_exists` fails without them. Unrelated to this change.

## Checklist

- [x] `python tests/run_tests.py --all` passes (activate `.venv` first), with a live Redmine configured — verified on 7.0.0 and 6.1.2
- [x] Change works with both local (`uv run python -m redmine_mcp_server.main`) and Docker (`docker-compose up`) deployments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   that replaces the row. Only a missing row (HTTP 404) takes the create path;
   any other error is surfaced, so a transient failure cannot null the agile
   fields it was unable to read.
+- `manage_redmine_wiki_page` no longer fails against Redmine 7.0 with
+  `'dict' object has no attribute 'id'`. Redmine 7.0 added a `project` ref to
+  the wiki page API response
+  ([Redmine #43569](https://www.redmine.org/issues/43569)), and python-redmine
+  hands that field over as a plain dict rather than a resource object, so
+  serializing it raised. This broke `get`, `create`, `update`, and `rename` —
+  the write actions changed the page and only then failed, reporting an error
+  for a change that had in fact been applied. The ref is now read in either
+  shape and those actions return the page again. `list` and `delete` were
+  unaffected, as are Redmine versions before 7.0, which omit the field.
 
 ### Contributors
 - @knasiotis — reported the dropped `agile_sprint_id` write and implemented
--- a/docs/tool-reference.md
+++ b/docs/tool-reference.md
@@ -1779,7 +1779,7 @@ List, get, create, update, delete, or rename a Redmine wiki page. Replaces `list
 
 **Returns:**
 - `list`: array of page metadata dicts (`title`, `version`, `parent_title` if present, `created_on`, `updated_on`) — no body text
-- `get`/`create`/`update`: full wiki page dict (`title`, `text`, `version`, `created_on`, `updated_on`, `author`, `project`, `attachments` when applicable). `project` is only returned by Redmine 7.0+ ([#43569](https://www.redmine.org/issues/43569)); earlier versions omit the key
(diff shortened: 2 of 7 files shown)


jztan (approved):
Thanks @azelcs. I rebuilt a 7.0.0 sandbox from a snapshot of my 6.1.1 data and hit exactly what you did: all four actions blow up on develop, all four come back with a real `project` ref on your branch, and the lifecycle test fails and passes to match. Unit suite green, black and flake8 clean.

The `_named_ref` fix is the part I would have missed. That silent `{"id": None, "name": ""}` was quietly eating data at six other call sites.

I will add your CHANGELOG credit myself at merge.

PR #221, opened 2026-08-20
fix: advertise CRM scopes so manage_contact can reach Redmine

### LINKED ISSUE #220: manage_contact is denied for every action in OAuth mode: CRM permissions are never advertised as scopes
### Bug Description

`manage_contact` is denied by Redmine for **every** action in `oauth` and `oauth-proxy` mode, even with `REDMINE_CRM_ENABLED=true`, the RedmineUP CRM plugin installed, the Contacts module enabled on the project, and a user whose Redmine role grants the CRM permissions.

The cause is that the CRM permissions are never advertised as scopes, so no access token can carry them — and Redmine intersects a token's scopes with the user's role permissions:

- `Role#allowed_permissions` (`app/models/role.rb`) is literally `unscoped & scope`, where `scope` is `@oauth_scope`.
- `ApplicationController#find_current_user` sets `user.oauth_scope = access_token.scopes.all.map(&:to_sym)` for any Doorkeeper-authenticated request.

Every contacts endpoint `manage_contact` calls then runs a scope-aware authorization check:

| Action | Gate |
|---|---|
| `list` | `contacts#index` → `find_optional_project` → `authorize_global` → `User#allowed_to?` |
| `get` | `contacts#show` → `find_contact` → `Contact#visible?` → `user.allowed_to?(:view_contacts, project)` |
| `create` | `contacts#create` → `before_action :authorize` |
| `update` / `delete` | `Contact#editable?` / `Contact#deletable?` |
| `assign_to_project` / `remove_from_project` | `contacts_projects#create` / `#destroy` → `Contact#editable?` plus an explicit `allowed_to?(:edit_contacts, ...)` |

Because `oauth_scopes.py` excludes vendor scopes and `advertised_scopes()` has no CRM branch, a token holds at most the core read/write scopes. `view_contacts` is absent, the intersection drops it, and the request is denied.

Worth separating from `TOOL_SCOPES["manage_contact"] = frozenset()`: that only means the MCP layer does not enforce a scope. The denial comes from Redmine itself, so per-tool enforcement being off does not help.

### Steps to Reproduce

1. Install the RedmineUP CRM plugin, enable the Contacts module on a project, and grant a user's role `view_contacts` plus `add_contacts` / `edit_contacts` / `delete_contacts`.
2. Run the server with `REDMINE_AUTH_MODE=oauth-proxy` (or `oauth`), `REDMINE_CRM_ENABLED=true`, `REDMINE_MCP_READ_ONLY=false`.
3. Register the Redmine OAuth application with the scopes this server advertises and complete a login as that user.
4. Call `manage_contact(action="list")`, then any other action.

### Expected Behavior

The contacts that user's role can see are returned, and the write actions succeed where the role permits, with Redmine's own role and visibility rules as the boundary.

### Actual Behavior

Redmine denies every one of the seven actions (`deny_access` / 403). The tool is unusable in either OAuth mode regardless of role.

### Version

`develop`; also present in 2.11.0.

### Deployment Method

Docker

### MCP Client

Hosted connector over `oauth-proxy`.

### Redmine Version

6.1.1, with RedmineUP CRM (`redmine_contacts`) 4.4.5.

### Additional Context

This is the same class of gap that `AGILE_READ_SCOPES` (#173) and `TAGS_READ_SCOPES` / `TAGS_WRITE_SCOPES` already fixed for their plugins: advertise the plugin permissions only when that plugin's feature flag is set, so the token can reach the endpoints while a deployment without the plugin never sees an unresolvable scope. A PR following that pattern is on the way.

`manage_product` looks like it has the same problem for the same reason — `TOOL_SCOPES` maps it to `frozenset()` and no product scopes are advertised — but I have not tested it against a Redmine with the Products plugin, so I have left it out rather than guess.

mmahmed:
## Summary

Fixes #220.

`manage_contact` is denied by Redmine for every action in `oauth` and `oauth-proxy` mode, because the RedmineUP CRM permissions are never advertised as scopes and Redmine intersects a token's scopes with the consenting user's role permissions (`Role#allowed_permissions` is `unscoped & scope`). No token could carry `view_contacts`, so all seven actions failed even for users whose role granted the permission.

This follows the existing `REDMINE_AGILE_ENABLED` / `REDMINE_TAGS_ENABLED` pattern: the plugin permissions live in their own lists and are advertised only when that plugin's feature flag is set, so a deployment without the plugin never advertises a scope its Redmine cannot resolve.

`TOOL_SCOPES` is deliberately left alone. The scopes are advertised so tokens can carry them, not *required* per tool — requiring them would gate the tool on a scope that a CRM-disabled deployment never advertises, and Redmine already enforces its own plugin permissions.

## Changes

- Add `CRM_READ_SCOPES` (`view_contacts`, `view_private_contacts`) and `CRM_WRITE_SCOPES` (`add_contacts`, `edit_contacts`, `delete_contacts`), each commented with the endpoint and the permission mapping that requires it.
- `advertised_scopes()` appends the read scopes when `REDMINE_CRM_ENABLED` is truthy, and the write scopes unless `REDMINE_MCP_READ_ONLY` is set.
- `view_private_contacts` is included because `Contact.visible_condition` consults it for `visibility=2` contacts; omitting it would silently narrow `list` / `get` results for a user whose role grants it — the same failure mode this PR fixes, one level down.
- `edit_contacts` covers `assign_to_project` and `remove_from_project`: the plugin maps `contacts_projects#create` / `#destroy` to it, and `ContactsProjectsController` also checks it directly.
- Update the module docstring's Exclusions note and the `TOOL_SCOPES` comment, which both claimed CRM scopes are never advertised.
- Docs: README env table, plugin table and gated-tool list; `docs/oauth-setup.md` scope-enforcement note; `docs/tool-reference.md` `manage_contact` prerequisites.
- 8 new tests covering the constants, flag gating across truthy/falsy/unset values, and read-only behaviour.

## Testing

```
python tests/run_tests.py --all
  1651 passed (unit)
  13 passed, 76 skipped (integration; skips need a live Redmine)

uv run black --check src/     All done! 39 files would be left unchanged.
uv run flake8 src/ --max-line-length=88     clean
```

`tests/test_ssl_integration.py` needs `tests/fixtures/ssl/generate-test-certs.sh` run first, otherwise one test fails on the missing fixture on a clean checkout too.

The live-Redmine integration tests are skipped here, so the end-to-end consent-and-call path against a Redmine with the CRM plugin has not been exercised in CI. `advertised_scopes()` gating is covered by unit tests.

## Checklist

- [x] No manual version bumps: `pyproject.toml`, `server.json`, and CHANGELOG version headers untouched
- [x] Docs updated where behavior changed: README stays concise, details in `docs/tool-reference.md` and `docs/oauth-setup.md`
- [x] Change works with both local and Docker deployments (config-only change, no new dependency or runtime requirement)

### Note for deployers

Enabling `REDMINE_CRM_ENABLED` now widens the advertised scope set. Those scopes must also be granted on the Redmine OAuth application, and existing users have to re-consent, since Doorkeeper freezes a token's granted scopes at issue time. Deployments that pin the advertised scope count in their own checks will see it change. Called out in the CHANGELOG.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,23 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- `manage_contact` no longer fails for every action in OAuth mode. Redmine
+  intersects an OAuth token's scopes with the consenting user's role
+  permissions (`Role#allowed_permissions` is `unscoped & scope`), and every
+  contacts endpoint the tool calls runs a scope-aware authorization check --
+  `contacts#index` through `authorize_global`, `contacts#show` through
+  `Contact#visible?`, and the mutations through `authorize`,
+  `Contact#editable?` and `Contact#deletable?`. Since the CRM permissions were
+  never advertised, no token could carry them and all seven actions were denied
+  even for users whose Redmine role granted the permission. Setting
+  `REDMINE_CRM_ENABLED=true` now adds `view_contacts` and
+  `view_private_contacts` to the advertised scopes, plus `add_contacts`,
+  `edit_contacts` and `delete_contacts` unless `REDMINE_MCP_READ_ONLY` is set,
+  following the existing `REDMINE_AGILE_ENABLED` / `REDMINE_TAGS_ENABLED`
+  pattern. Deployments enabling the flag must grant the new scopes on the
+  Redmine OAuth application and have users re-consent.
+
 ### Tests
 - Unit coverage for `_auth.py`, raised from 63% to 100%. The whole
   `revoke_token` RFC 7009 proxy was untested, along with the fail-fast branch
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ The server runs on `http://localhost:8000` with the MCP endpoint at `/mcp`, heal
 | `REDMINE_AGILE_ENABLED` | No | `false` | Enable RedmineUP Agile plugin support: `get_redmine_issue` returns `story_points`, `agile_sprint_id`, `agile_position`; `update_redmine_issue` accepts `story_points` |
 | `REDMINE_CHECKLISTS_ENABLED` | No | `false` | Enable RedmineUP Checklists plugin support: `get_checklist`, `create_checklist_item`, `update_checklist_item` (requires Checklists Pro plugin) |
 | `REDMINE_PRODUCTS_ENABLED` | No | `false` | Enable RedmineUP Products plugin support: `manage_product` (action=list/get/create/update) |
-| `REDMINE_CRM_ENABLED` | No | `false` | Enable RedmineUP CRM plugin support: `manage_contact` (action=list/get/create/update/delete/assign_to_project/remove_from_project) |
+| `REDMINE_CRM_ENABLED` | No | `false` | Enable RedmineUP CRM plugin support: `manage_contact` (action=list/get/create/update/delete/assign_to_project/remove_from_project). Requires the CRM plugin and the `view_contacts` / `view_private_contacts` permissions on the Redmine server, plus `add_contacts` / `edit_contacts` / `delete_contacts` for the write actions. In OAuth mode these are advertised as scopes only when this flag is set, so the OAuth application must grant them too. |
 | `REDMINE_DMSF_ENABLED` | No | `false` | Enable DMSF document-management plugin support: `manage_document` (action=list/get/create/update). Requires `redmine_dmsf` plugin on the Redmine server. |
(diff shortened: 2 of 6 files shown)


jztan:
Thanks @mmahmed. On a 6.1.1 sandbox with the Light `redmine_contacts` plugin (v4.4.7), Contacts module enabled, and CRM permissions granted to a non-admin role: a token without the CRM scopes gets 403 on every `/contacts.json` verb, a token with them gets through.

| Action | Token without CRM scopes | Token with CRM scopes |
|---|---|---|
| list | 403 | 200 |
| create | 403 | 201 |
| update | 403 | 204 |
| delete | 403 | 204 |

Matches the mechanism you described: Redmine intersects the token's scopes with the role's permissions, so a role grant alone isn't enough. `advertised_scopes()` on `develop` never includes the CRM scopes under `REDMINE_CRM_ENABLED=true`; your branch adds `view_contacts` / `view_private_contacts` / `add_contacts` / `edit_contacts` / `delete_contacts`, dropping the three write scopes when `REDMINE_MCP_READ_ONLY=true`.

Merging this now. Thanks for the writeup, and for leaving `manage_product` out rather than bundling in a fix you hadn't verified.

PR #223, opened 2026-08-20
fix: read issue includes from the payload instead of re-fetching

### LINKED ISSUE #222: include=relations is discarded and re-fetched per issue, requiring manage_issue_relations
### Bug Description

`get_redmine_issue(include_relations=True)`, `get_gantt_chart` and `delete_redmine_issue` all ask Redmine for `include=relations` and then read the result with `getattr(issue, "relations")`. That attribute never reads the include payload.

python-redmine lists `relations` in **both** `Issue._includes` and `Issue._relations` (`redminelib/resources/standard.py`), and `BaseResource.__getattr__` tests `_relations` first, so the attribute returns a lazy `ResourceSet` that issues `GET /issues/{id}/relations.json` on first iteration.

Redmine maps that endpoint's `index` action to `manage_issue_relations` (`lib/redmine/preparation.rb`, inside `map.project_module :issue_tracking`), a permission that reading an issue does not imply. It also carries no `:read => true` flag, so it is withheld in closed and archived projects. So a caller holding `view_issues` is denied relations Redmine had already put in the payload, and every caller pays an extra round trip.

### Steps to Reproduce

1. Give a user `view_issues` but **not** `manage_issue_relations` on a project.
2. Call `get_redmine_issue(issue_id=N, include_relations=True)` for an issue in that project that has at least one relation.
3. Observe a permission error rather than the relations, even though the response to the single `GET /issues/N.json include=relations` already contained them.

To see the extra request without needing a restricted role, count requests with a fake transport under real python-redmine, making the endpoint return different data than the payload so the source being read is unambiguous:

```python
issue = redmine.issue.get(1, include="relations")

# [after get]               1 request   GET /issues/1.json?include=relations
# [after `issue.relations`] 1 request   (lazy ResourceSet, nothing sent yet)
# [after iterating it]      2 requests  GET /issues/1/relations.json?issue_id=1

payload relations ids = [11, 12]
issue.relations ids   = [999]      # came from the network
raw()['relations']    = [11, 12]   # the include was there the whole time
```

For the `delete_redmine_issue` path, call it on a **leaf** issue (no subtasks) and count requests:

```
GET /issues/1.json?include=journals,attachments,relations,children   <- the real fetch
GET /issues/1.json?include=children                                  <- leaf-issue refetch
GET /issues/1/relations.json                                         <- relations refetch
GET /time_entries.json?issue_id=1
4 requests
```

### Expected Behavior

One request serves the whole call. `include=relations` is honoured from the response payload, so `view_issues` is sufficient, `get_gantt_chart` costs one request per chart rather than one per issue, and the `delete_redmine_issue` preview does not make hidden calls that can fail before its confirmation gate.

An unreadable time-entry count should be reported as unknown rather than `0`, since `0` understates an irreversible cascade.

### Actual Behavior

| Tool | Effect |
|---|---|
| `get_redmine_issue(include_relations=True)` | denied without `manage_issue_relations`; the include is paid for and discarded |
| `get_gantt_chart` | one extra request **per issue** (it already requests `include=relations`), plus the same denial |
| `delete_redmine_issue` | same denial, plus an extra request per leaf issue, and both are uncaught |

`delete_redmine_issue` is the worst case. Its impact preview reads all four included collections through resource attributes, and those reads sit **outside** the `try` that wraps the fetch, so any request they trigger escapes as an unhandled exception rather than the tool's normal error envelope — before the confirmation gate a caller relies on. Two of the four really do fire a request:

- `relations`, always, for the reason above.
- `children`, for every **leaf** issue: Redmine's `render_api_issue_children` starts with `return if issue.leaf?`, so the key is absent from the payload, and python-redmine re-fetches whenever an include key is missing.

The comment above that block states the intent the code does not keep — that the includes "let us populate a preview without separate API round-trips", and that an absent `children` key should simply count as 0.

`time_entries` is a genuinely separate case: it is also in `Issue._relations`, but Redmine has no `time_entries` include for an issue at all, so that request is unavoidable. It needs `view_time_entries`, and a denial there is currently reported as `time_entries_count: 0`.

### Version

2.11.0 (`develop` @ `ff1440d`)

### Deployment Method

Source (development)

### MCP Client

Other (specify in additional context)

### Redmine Version

6.1.1

### Additional Context

Reproduced directly against the tool functions with an instrumented HTTP transport rather than through a particular MCP client, since the behaviour is in the server's use of python-redmine (2.5.0) and is client-independent.

Suggested fix: read these collections from the payload python-redmine already decoded (`raw()`) rather than through the lazy attribute. That removes a permission requirement instead of adding a scope, removes the Gantt N+1, and makes the delete preview cost one request. For `time_entries`, keep the request but report an unreadable count as unknown.

Happy to send a PR — I have one ready with tests if that helps.

mmahmed:
## Summary

`include=relations` was requested and then thrown away. python-redmine lists `relations` in both `Issue._includes` and `Issue._relations`, and `BaseResource.__getattr__` tests `_relations` first, so `getattr(issue, "relations")` returns a lazy resource set that fetches `GET /issues/{id}/relations.json` instead of reading the payload. Redmine maps that endpoint's `index` to `manage_issue_relations` — a permission reading an issue does not imply, and one with no `:read => true` flag, so it is also withheld in closed and archived projects. Callers holding `view_issues` were
denied relations Redmine had already sent them.

Fixes #222

## Changes

- Add `_included_list`, which reads an include from python-redmine's decoded payload via `raw()`, and move `_issue_relation_to_dict` beside it so a single serializer covers both payload dicts and `IssueRelation` resources.
- Use the payload in `get_redmine_issue`, `get_gantt_chart` — one request per chart instead of one per issue — and the `delete_redmine_issue` preview.
- Fix the whole delete preview, not just relations. All four included collections were read through resource attributes outside the enclosing `try`, so any request they triggered escaped as an unhandled exception before the confirmation gate. Two really fired: `relations` always, and `children` for every leaf issue, since `render_api_issue_children` returns early on `issue.leaf?` and python-redmine re-fetches a missing include. A leaf-issue preview drops from four requests to two.
- Report `time_entries_count` as `null` when it cannot be read. Redmine has no `time_entries` include, so that request is unavoidable and needs `view_time_entries`; `0` would understate an irreversible cascade. Other failures return a normal error envelope rather than a permission diagnosis.
- Relations now carry Redmine's `delay` field on the single-issue path too, which it had omitted while the Gantt path included it.

One deliberate test change: the relation fakes in `tests/test_gantt.py`, `tests/test_delete_redmine_issue.py` and `tests/test_redmine_handler.py` served includes as plain attributes, which no python-redmine issue does. They passed while the real client made an extra request that could 403 — which is how this survived. They now serve includes from `raw()` and raise on the attribute, so a regression fails in CI rather than only against a live Redmine. No assertion was weakened or removed.

## Testing

`python tests/run_tests.py --all` — passes.
`uv run black --check src/` and `uv run flake8 src/ --max-line-length=88` — clean.

New `tests/test_relations_payload.py` (12 tests) covers the payload read, an absent include, an empty include, one request for a whole Gantt chart, the leaf-issue `children` case, the preview surviving a caller without `manage_issue_relations`, an unreadable time-entry count reported as unknown rather than zero, and an unexpected time-entry error returning an error envelope. Each was checked to fail against `develop` before the fix.

Verified end to end by counting requests through a fake transport under real python-redmine, on a leaf issue:

```
before: 4 requests   issue.json(include=...) + issue.json(include=children)
                     + issues/1/relations.json + time_entries.json
after:  2 requests   issue.json(include=...) + time_entries.json
```

With the relations endpoint deliberately returning different data than the payload, the output now matches the payload — confirming the include is what is read, not a second request.

## Checklist

- [x] No manual version bumps
- [x] Docs updated where behavior changed: `docs/tool-reference.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,37 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Fixed
+- `include=relations` is no longer discarded and re-fetched per issue.
+  python-redmine lists `relations` in both `Issue._includes` and
+  `Issue._relations`, and `BaseResource.__getattr__` tests `_relations` first,
+  so `getattr(issue, "relations")` ignored the payload the `include` had
+  already fetched and handed back a lazy resource set that requested
+  `GET /issues/{id}/relations.json` as soon as it was iterated. Redmine maps
+  that endpoint's `index` action to `manage_issue_relations`, a permission
+  reading an issue does not imply -- and one carrying no `:read => true` flag,
+  so it is also withheld in closed and archived projects. Callers holding
+  `view_issues` were therefore denied relations Redmine had already sent them.
+  Relations now come from the response payload: `get_redmine_issue(
+  include_relations=True)` works with `view_issues` alone, and
+  `get_gantt_chart` drops from one extra request per issue to none
+  ([#222](https://github.com/jztan/redmine-mcp-server/issues/222)). Each
+  relation now also carries Redmine's `delay` field, which the single-issue
+  path had omitted.
+- `delete_redmine_issue` no longer makes unguarded extra requests while
+  building its impact preview. All four included collections were read through
+  resource attributes rather than the payload, outside the `try` that wraps the
+  fetch, so the request each one could trigger escaped as an unhandled
+  exception instead of an error envelope -- before the confirmation gate. Two
+  of them really did fire: `relations` always, and `children` for every leaf
+  issue, because Redmine omits that key entirely (`render_api_issue_children`
+  returns early on `issue.leaf?`) and python-redmine re-fetches a missing
+  include. All four now come from the payload, which is what the code's own
+  comment always said the preview did. `time_entries` is the one count Redmine
+  has no issue include for, so it still needs its own request: a caller without
+  `view_time_entries` now sees `time_entries_count: null` rather than `0`,
+  which would have understated an irreversible cascade, and any other failure
+  returns a normal error envelope instead of being reported as a permission
+  problem.
 - `manage_contact` no longer fails for every action in OAuth mode. Redmine
   intersects an OAuth token's scopes with the consenting user's role
   permissions (`Role#allowed_permissions` is `unscoped & scope`), and every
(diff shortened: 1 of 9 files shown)


jztan:
Thanks @mmahmed, merged. Nice catch on the relations re-fetch, that 403-on-view-only-permission bug would've bitten anyone without manage_issue_relations.

Issue #109, opened 2026-05-15
Stop wrapping structured metadata fields (filename, author.name) in <insecure-content>


jztan:
## Problem

Prompt-injection wrapping is the right design for free-text fields (descriptions, journals, notes). For **structured metadata** it's over-defensive and adds parsing friction:

- `attachment.filename` is wrapped, but a filename is structured. Callers must strip the wrapper before passing to `upload_file`, displaying, or saving to disk.
- `author.name` is wrapped, same issue.

## Proposal

Wrap only free-text: \`description\`, \`notes\`, journal content, free-text custom-field values. Leave bare: filenames, names, emails, timestamps, IDs.

Source: LLM-driven evaluation, 2026-05-15.

jztan:
Fixed in 9fffe8b (on \`develop\`).

**Policy shift:** wrap based on **who authored the field**, not on what schema slot it sits in.

**STAY WRAPPED** (free-text the user authored as prose):
\`description\`, \`notes\`, journal content, time-entry \`comments\`, wiki page \`text\`, search \`excerpt\`, attachment \`description\`, version/product/contact \`description\`/\`background\`, checklist item \`subject\` (looks like a label but is user-authored prose -- the classic place injection hides).

**VERBATIM** (structured metadata):
filenames (issue attachments, project files, upload responses, \`get_redmine_attachment\`), all \`_named_ref\` display names (author, version, project, tracker, status, role, etc.), contact name fields (\`first_name\`, \`last_name\`, \`middle_name\`, \`company\`, \`job_title\`, \`assigned_to.name\`), product \`name\`/\`code\`/\`project.name\`/\`category.name\`, issue category \`name\`, gantt issue \`subject\` and version \`name\`, project member role \`name\`.

**Why this matters:** \`<insecure-content-{nonce}>\` wrappers on names and filenames were the single most common reason agents had to post-process MCP responses before using them. "Pass this filename to another tool" and "format this author for display" flows just got simpler.

**Backward-incompatible** for any caller that explicitly stripped wrapper tags from these fields -- such callers can simplify.

**Drift guard:** \`tests/test_wrap_policy.py\` has two fixtures (\`TestStructuredFieldsAreVerbatim\`, \`TestFreeTextFieldsRemainWrapped\`) so accidental drift fails CI from whichever direction it happens. The full surface was sampled in verification (see PR / commit thread) -- no leakage observed in either direction.

[...]

PR #225, opened 2026-08-20
feat: add manage_deal for RedmineUP CRM deals

### LINKED ISSUE #224: Add a manage_deal tool for RedmineUP CRM deals
## Problem Statement

The RedmineUP CRM plugin exposes two resources over its REST API, contacts and deals. `manage_contact` covers contacts. Nothing covers deals, so a pipeline is only reachable by calling `/deals.json` directly, outside this server.

Deals are the half of the plugin that carries the commercial state — amount, currency, probability, close date, stage — so the gap tends to land on exactly the questions a CRM gets asked.

There is a second, less obvious half. Even with a tool, deals would be unreachable in OAuth mode: `view_deals`, `add_deals`, `edit_deals` and `delete_deals` are not advertised, so no token can carry them. Redmine intersects a token's scopes with the consenting user's role permissions (`Role#allowed_permissions` is `unscoped & scope`), which is the same mechanism behind #220. Every action is denied, by two different routes: `deals#show`, `#create`, `#update` and `#destroy` through `before_action :authorize`, and `deals#index` through `find_optional_project`, which Redmine implements as a permission check calling `authorize_global`. `Deal.visible` then narrows what a permitted caller sees; it is not what gates access.

## Proposed Solution

Add `manage_deal`, shaped like `manage_contact`: one tool, action-dispatched, with actions `list`, `get`, `create`, `update`, `delete` — the five verbs the plugin's `DealsController` declares in `accept_api_auth`. There is no equivalent of `assign_to_project` / `remove_from_project`: a contact is many-to-many with projects, a deal has a single `project_id`.

Response keys mirror the plugin's own `app/views/deals/index.api.rsb` and `show.api.rsb` rather than introducing new names: `id`, `name`, `price`, `currency`, `price_type`, `duration`, `probability`, `due_date`, `background`, `project`, `status`, `category`, `author`, `contact`, `assigned_to`, `related_contacts`, `created_on`, `updated_on`. Writable fields mirror `Deal`'s `safe_attributes`.

Advertise `view_deals`, and `add_deals` / `edit_deals` / `delete_deals` unless the server is read-only, following the same shape as the existing CRM scopes. `view_deals` is declared `:read => true` in the plugin, so it also applies in closed and archived projects.

Two behaviours of the plugin are worth surfacing in the tool rather than leaving for callers to discover. `DealQuery` seeds itself with a `status_id` filter on operator `o`, so an unfiltered `list` silently returns only open deals — `status_id` is a `:list_status` filter, so accepting `"*"` / `"c"` / a numeric id alongside the default gives callers a way out. And `DealsController#create` runs `price` through `parsed_price`, which calls `String#gsub!`, so a JSON number raises `NoMethodError` server-side; the price has to go over the wire as a string.

**Gate them behind a new `REDMINE_DEALS_ENABLED` rather than the existing `REDMINE_CRM_ENABLED`.** This is the one place the change departs from the project's one-flag-per-plugin convention, and the reason is that deals are absent from the CRM plugin's Light edition: it declares no `project_module :deals` and therefore none of the `*_deals` permissions. Redmine builds its OAuth scope list from `Redmine::AccessControl.permissions` and applies `enforce_configured_scopes`, so on a Light install those names are not valid scopes at all — the OAuth application cannot hold them, and a client requesting one fails consent with `invalid_scope`. Advertising deals under `REDMINE_CRM_ENABLED` would therefore break `manage_contact` for Light deployments that have that flag set and working today. A separate flag leaves them untouched.

Two things deliberately left out, both because they cannot work for a non-admin caller:

- **Deal statuses.** `DealStatusesController` has `before_action :require_admin, except: :assign_to_project`, so `GET /deal_statuses.json` is admin-only. A tool action for it would be dead for any deployment that withholds the `admin` scope. Each deal already carries its own `status` as an `{id, name}` ref, so reading is unaffected; only enumerating all statuses is out of reach.
- **Deal categories.** Reachable (`deal_categories#index` maps to `view_deals`), but per-project, and a deal carries its `category` ref already. Better as a follow-up than bundled here.

Consistent with `manage_contact`, `manage_product` and `manage_document`, `custom_fields` is accepted on writes but not returned on reads. Worth naming explicitly since the deals API does return it.

## Use Case

Reading a sales pipeline through MCP: list the deals on a project, then answer questions that need the commercial fields — total value by stage, which deals close this quarter, which are unassigned.

Also the write direction: advancing a deal's stage after a call, or creating a deal against an existing contact. Today both require leaving the MCP server for raw REST.

## Feature Category

New MCP Tool

## Priority

Important

## Alternatives Considered

**One flag: advertise the deal scopes under `REDMINE_CRM_ENABLED`.** This is what the one-flag-per-plugin convention would suggest, and it was the first design. It has to be rejected because of the Light edition: a deployment running CRM Light with `REDMINE_CRM_ENABLED=true` works today for contacts, and would start advertising deal scopes that Redmine cannot resolve, failing consent and taking contacts down with it. Already-issued tokens would survive, since Doorkeeper freezes granted scopes at issue time, but every new consent would fail. Documentation cannot prevent that, which is what makes the separate flag worth the deviation.

**Call `/deals.json` directly and skip the tool.** Works, but it needs credentials outside the server's auth model, and in OAuth mode it still fails: the token cannot carry `view_deals` no matter how the Redmine roles are set, because the scope is never advertised. So the scope half of this is required whether or not the tool is added.

**Advertise the deal scopes without adding the tool.** Would unblock a direct API caller but leaves the server with a visible gap in a plugin it otherwise supports.

## Additional Context

Field names, permission mappings and controller behaviour above were read from the plugin's own `init.rb`, `DealsController`, `Deal` model and API view templates at CRM **4.2.6 (PRO)**, and the response field list was cross-checked against the vendor's current published REST API documentation for deals, which matches.

The Light/Pro split was established by comparing that source against CRM **4.3.6** and **4.4.1** ("Light version"): neither declares `project_module :deals`, and `view_deals` and `add_deals` appear nowhere in either. The Redmine side is `config/initializers/30-redmine.rb` on 6.1-stable, which sets `optional_scopes` from `Redmine::AccessControl.permissions` and enables `enforce_configured_scopes`.

One documentation discrepancy worth flagging upstream of this project: the vendor's published deal-statuses API page does not mention the admin requirement that `DealStatusesController` enforces.

## Contribution

mmahmed:
## Summary

Adds `manage_deal`, a tool for RedmineUP CRM deals, and advertises the deal permissions so a token can actually carry them. Fixes #224.

Deals were the one CRM resource with no tool, so reaching a pipeline meant calling `/deals.json` outside this server. The scope half matters independently: `view_deals`, `add_deals`, `edit_deals` and `delete_deals` were never advertised, and Redmine intersects a token's scopes with the consenting user's role permissions, so the endpoints were unreachable at any role. That is the same mechanism as #220.

All five actions are denied without the permission, by two routes: `deals#show`, `#create`, `#update` and `#destroy` through `before_action :authorize`, and `deals#index` through `find_optional_project`, which Redmine documents as "Find a project based on params[:project_id] and authorize the user for the requested action" and implements with `authorize_global`. `Deal.visible` narrows what a permitted caller sees; it does not stand in for the check.

## Changes

- New `src/redmine_mcp_server/tools/deals.py` with `manage_deal`, action dispatched over `list`, `get`, `create`, `update`, `delete` — the five verbs `DealsController` declares in `accept_api_auth`. No `assign_to_project` / `remove_from_project`: a deal has a single `project_id`, unlike a contact.
- Response keys mirror the plugin's own `app/views/deals/index.api.rsb` and `show.api.rsb` rather than introducing new names, and writable fields mirror `Deal`'s `safe_attributes`. A test pins the writable set so any edit to it has to be deliberate — it restates the constant, so it cannot detect drift in the plugin itself.
- `price` is sent as a string. `DealsController#create` runs it through `parsed_price`, which calls `String#gsub!`, so a JSON number raises `NoMethodError` and the request 500s on the obvious happy path. Non-finite floats and booleans are rejected rather than emitted as bare `NaN` / `true`.
- `list` documents and works around `DealQuery`'s seeded `status_id` filter on operator `o`, which otherwise makes an unfiltered call silently return open deals only. `status_id` accepts `"*"`, `"c"` or a numeric id.
- Ids supplied through `fields` are validated the same way the named parameters are. Without that, `fields={"status_id": True}` reaches Redmine as `true` and ActiveRecord casts it to `1`, selecting record 1 instead of failing — the exact case `_is_positive_int` exists to prevent.
- `notes` are serialized when `include="notes"` returns them, keyed as `show.api.rsb` renders them, with `content` wrapped as free text. The key is omitted when not requested, so absence stays distinguishable from "none present".
- `deals` added to `get_mcp_server_info`'s `plugin_flags`, and to the pinned set in `tests/test_server_info.py`. Without it `manage_deal` would be the only plugin-gated tool a caller cannot pre-flight, which `docs/tool-reference.md` tells them to do.
- `background` is wrapped with `wrap_insecure_content`; `name` and the `{id, name}` refs are returned verbatim, matching how `subject` and the existing refs are handled per #109. Refs reuse `_named_ref`, and the plugin omits a ref entirely when the association is nil, so absent refs serialize as `None`.
- New `DEALS_READ_SCOPES` (`view_deals`) and `DEALS_WRITE_SCOPES` (`add_deals`, `edit_deals`, `delete_deals`), advertised under a new `REDMINE_DEALS_ENABLED` flag, with writes withheld in read-only mode. Each entry carries a comment naming the controller action it unlocks, and the module docstring's exclusions index lists them.
- `TOOL_SCOPES` gains `manage_deal: frozenset()`, matching `manage_contact`: Redmine enforces the plugin's permissions, and requiring them here would gate the tool on a scope a deals-disabled deployment never advertises.
- `manage_deal` added to the `EXPECTED_ACTIONS` and `EXPECTED_BOUNDS` registries in the existing schema tests.
- Docs: new `manage_deal` section in `docs/tool-reference.md`, plus its rows in that file's read-only breakdown; tool counts moved everywhere they appear — `README.md` (51 → 52, max 52 → 53, plugin-gated 6 → 7), `docs/contributing.md` (including a `tools/deals.py` row in the module table and the tree comment) and `docs/roadmap.md`; `REDMINE_DEALS_ENABLED` documented in the README env-var and plugin tables, `docs/oauth-setup.md`, `.env.example` and `.env.docker.example`; read-only action list in `docs/troubleshooting.md` extended; CHANGELOG `[Unreleased]` entry.

**On the new flag, since it deviates from one-flag-per-plugin.** Deals ship inside the CRM plugin, so `REDMINE_CRM_ENABLED` would be the conventional gate, and that was the first design. It cannot be used, because deals are absent from the plugin's Light edition: it declares no `project_module :deals` and therefore none of the `*_deals` permissions. Redmine derives its OAuth scope list from `Redmine::AccessControl.permissions` and applies `enforce_configured_scopes` (`config/initializers/30-redmine.rb`), so on a Light install those names are not valid scopes — the OAuth application cannot hold them and a client requesting one fails consent with `invalid_scope`. Advertising deals under the CRM flag would therefore break `manage_contact` for Light deployments that have that flag set and working today; already-issued tokens would survive, but every new consent would fail. A separate flag leaves those deployments completely untouched.

Deal statuses are deliberately not exposed. `DealStatusesController` has `before_action :require_admin, except: :assign_to_project`, so `GET /deal_statuses.json` cannot serve a non-admin token, and an action for it would be dead in any deployment that withholds the `admin` scope. Each deal carries its `status` as an `{id, name}` ref, so reading a deal's stage is unaffected; the tool's docstring and the tool reference say where to read status and category IDs from instead. Deal categories are reachable but per-project, and are left for a follow-up. `custom_fields` is accepted on writes but not returned on reads, matching `manage_contact`, `manage_product` and `manage_document`.

Nothing existing changes: no signature, response key, or previously advertised scope moves, and a deployment that does not set the new flag sees no difference.

## Testing

- `python tests/run_tests.py --all` — exits 0.
- `python -m pytest -m "not integration" tests/` — 1745 passed, 89 deselected, 0 failures. Integration tests skip without a live Redmine configured.
- `black --check src/` and `flake8 src/ --max-line-length=88` — both clean.
- New `tests/test_deals_support.py`, 92 tests: scope advertising across the enabled, read-only and disabled combinations, including that `REDMINE_CRM_ENABLED` alone advertises no deal scopes, that the two flags are independent in both directions, and that no deal scope leaks into the always-on lists; the flag accessor itself; read-only mode per write action; serialization, including verbatim `name`, wrapped `background`, absent refs becoming `None`, timestamps, and notes; price coercion and the values that are refused; id validation through `fields`; error envelopes for a transport failure on every action; and each action's success path, request shape and validation rejections (including a boolean `deal_id`, which is an `int` subclass in Python and must not read deal 1).
- Each load-bearing assertion was confirmed to fail before the corresponding change, by mutation: the scope-advertising tests against `develop`; the `REDMINE_CRM_ENABLED`-alone guard against the single-flag version of `advertised_scopes()`; `test_every_registered_tool_is_mapped` without the `TOOL_SCOPES` entry; the read-only tests with the three write actions flipped to `ActionMode.READ`; the price tests with the string coercion removed; and the delete test with the URL path changed.
- Tool registration verified by listing tools: 52 registered, 7 plugin-gated, 45 core, matching the README counts this PR updates.

Not exercised against a live RedmineUP CRM install with the deals module enabled; request shapes and field names come from the plugin source at 4.2.6 (PRO) and the vendor's current REST API documentation, which agree. The Light/Pro difference was established against 4.3.6 and 4.4.1. Not exercised in Docker either — the change adds no dependency and touches no build file, so the image is unaffected, but I have not run it there and have left that checklist box unticked rather than claim otherwise.

## Checklist

- [x] Change works with both local (`uv run python -m redmine_mcp_server.main`) and Docker (`docker-compose up`) deployments — local only; see Testing


jztan:
Thanks @mmahmed, merged. Good call flagging the docs-vs-source gap on the price field honestly rather than papering over it, and the separate REDMINE_DEALS_ENABLED flag reasoning around Light-edition OAuth consent was sharp.

PR #227, opened 2026-08-21
feat: return the contact fields the CRM API sends, and filter on them

### LINKED ISSUE #226: `manage_contact` drops contact fields the CRM API returns, and discards eight of its own `list` parameters
`manage_contact` builds its output from a fixed dict of key names, and several of those names differ from the ones the RedmineUP CRM API actually sends. The affected fields come back empty for every contact no matter how populated the record is, and three fields are dropped outright. Separately, `list` accepts eight parameters that its handler never forwards, so filtering on them silently returns every contact.

Verified against a Redmine running RedmineUP CRM, reading the raw API responses alongside the tool output.

## Fields the API sends and the tool does not return

`custom_fields` is not serialized at all. The CRM API renders contact custom field values on both `GET /contacts.json` and `GET /contacts/{id}.json`. The tool has no key for them, so a deployment that keeps anything meaningful in a contact custom field cannot read it back. The same values are writable: `custom_fields` is in `_CONTACT_WRITABLE_FIELDS`, so a contact custom field can be written through `manage_contact` and never read through it.

`author` is rendered on every contact and dropped.

`projects` is rendered on the single-contact route and dropped.

## Fields read under a key the API does not use

`email` and `phone` read the scalar keys `contact.get("email")` and `contact.get("phone")`. The API renders these as arrays — `emails` as `[{"address": ...}]` and `phones` as `[{"number": ...}]` — so both fields are `null` for every contact, including contacts that have an address on file.

`tags` reads `contact.get("tags")`. The API sends `tag_list`, so `tags` is always `[]`.

`address` is rebuilt from a fixed six-key set including `street1` and `street2`. The API's address sub-document uses `street` and also carries `full_address`, so the tool reports `null` for two keys the plugin never sends and drops one it does.

## `list` parameters that are accepted and discarded

`_list_contacts_action` forwards `project_id`, `search`, `tags`, `assigned_to_id` and `limit`, and ends its signature with `**_: Any`. The tool's schema also offers `first_name`, `last_name`, `company`, `email`, `phone`, `is_company`, `visibility` and `include`. On `list` all eight are swallowed, so a caller filtering by name or company gets the unfiltered collection back and nothing indicates the filter was ignored.

Redmine does apply most of these server-side. Checking each candidate with a value that matches nothing — an applied filter answers with an empty collection, one Redmine does not recognise answers HTTP 200 with the whole collection — `first_name`, `last_name`, `middle_name`, `company`, `job_title`, `email`, `phone`, `tags`, `assigned_to_id` and `author_id` are all applied, as is `cf_<id>` for a custom field that has "Used as a filter" enabled. `address` and `visibility` are silently discarded and `last_note` rejects a bare value with 422, so those three are not worth offering.

There is also no `offset` parameter anywhere in the tool and `limit` is capped at 100, so a CRM with more than 100 contacts cannot be paged.

## Two parameters that should stay off the query

`is_company` reads like a filter — the schema types it `bool` with a default of `False` — but the plugin's filter is broken. Every spelling of it returns the same result set: `1`, `0`, `true`, `false`, and the explicit `f[]`/`op[]`/`v[]` form all return the same count, which does not match either side of a real company/person split. Forwarding it would answer a request for companies with people, so filtering on the `is_company` key in the response is the only reliable option.

`fields` must not be forwarded. Redmine reads `params[:fields]` (and its `f` alias) as the list of fields a query filters on, and that branch clears the query's filters before rebuilding them from the request. Passing it through therefore drops every other filter. This is easy to hit here rather than theoretical: `manage_contact` has its own `fields` parameter carrying create and update attributes, and the action dispatcher hands every tool parameter to every action, so `fields` already arrives at the list handler on every call and is currently saved only by the `**_` that swallows it. Confirmed on a live instance — a custom field filter that returns a narrow set on its own returns the full collection as soon as `fields` is added alongside it.

## Expected

Serialize the fields the API sends, read each under the key the API uses, forward the `list` filters Redmine applies, and add `offset`.

`custom_fields` needs no new permission or scope: Redmine has already limited the values to those the caller may see, the same as it does for issues and projects.

mmahmed:
Fixes #226.

The contact serializer built a fixed dict from key names that differ from the plugin's, so several fields were empty for every contact however populated the record was, and three were dropped outright. `list` separately accepted eight of its own parameters and discarded them.

## Serializer

- `custom_fields` is now returned. The CRM API renders contact custom field values on both routes and the tool had no key for them, so a value written through `fields` could never be read back.
- `author` and `projects` are returned. `projects` is emitted only when the payload carries it — the API sends it for a single contact and not for a collection — so an absent key is not reported as an empty list.
- `email` and `phone` read the `emails` and `phones` arrays the API sends. Both are returned as arrays, with the singular keys holding the first entry so existing callers keep working, and a payload that does use the scalar spelling is read exactly as before.
- `tags` reads `tag_list`, falling back to `tags`.
- `address` is passed through as the plugin names it, rather than rebuilt from a fixed key set that reported `null` for `street1`/`street2` and dropped `full_address`.

`_custom_fields_to_list` moves to `_serialization.py` and accepts a decoded payload mapping as well as a resource, since the contacts tool reaches the plugin endpoints through `engine.request` and holds JSON rather than a resource object. The issue serializer imports it from there and is otherwise unchanged.

## `list` filters

`first_name`, `last_name`, `middle_name`, `company`, `job_title`, `email`, `phone` and `author_id` are forwarded, and a `filters` dict carries anything this signature does not name — most usefully `cf_<id>` for a contact custom field. `offset` is added so a CRM past 100 contacts can be paged at all.

Each filter was checked against a running instance with a value that matches nothing, so an applied filter is distinguishable from one Redmine does not recognise: an applied filter answers with an empty collection, an unrecognised one answers HTTP 200 with the whole collection. `address` and `visibility` fall in the second group and `last_note` rejects a bare value with 422, so none of the three is offered.

Filters are built from the named parameters rather than from the collected keyword arguments, and `filters` refuses `fields` and `f`. Redmine reads those as the list of fields a query filters on and empties the query's filters before rebuilding them, so either one drops every other filter — reachable here rather than theoretical, because `manage_contact` carries create and update attributes in its own `fields` parameter and the dispatcher hands every parameter to every action.

`is_company` stays create-only. The plugin's filter returns the same result set for `1`, `0`, `true`, `false` and the explicit `f[]`/`op[]`/`v[]` form, matching neither side of a real company/person split, so forwarding it would answer a request for companies with people. The docstring says so and points at the `is_company` key in the response instead.

The docstring also gains an `Args` section. One flat signature serves all seven actions and nothing previously stated which parameter belonged to which, so `is_company` read like a filter and `fields` read like it applied to `list`.

## Notes

No new scope or permission. Redmine limits custom field values to those the caller may see before rendering them, the same as for issues and projects.

`include` on `get` is unchanged. On the instance used here no `include` value adds a key to the response, so there was nothing to serialize behind it; the keys the API does render unconditionally are what this change picks up.

## Validation

`tests/test_contact_payload_keys.py` is new. It covers the serializer against the shape the API actually returns, that a payload using the scalar `email`/`phone` spelling is unchanged, and that `projects` is absent rather than empty when the API did not send it. The filter tests cover what reaches Redmine and, more importantly, what must not: the tool's own `fields` parameter, an explicit `fields` or `f` inside `filters`, and `is_company`.

Full suite, `flake8` and `black` all pass.


jztan:
@mmahmed thanks for this, the field-mapping half is a real fix. I ran a create/get/list round trip against my sandbox (`redmine_contacts` 4.4.7, Light edition) and confirmed `email`/`phone` now read correctly from the `emails`/`phones` arrays, and `author`/`projects`/`custom_fields` all come back populated instead of empty.

The filter half doesn't hold up on that same sandbox. I created two contacts with different `company`, `first_name`, `last_name`, `email`, `job_title`, and `phone` values, then queried `list` with each of those six filters set to a value that matches neither contact. Every one of them returned both contacts instead of an empty list, which contradicts the comment in `_list_contacts_action` claiming a non-matching value comes back empty.

I traced it to `ContactQuery#initialize_available_filters` in this plugin version: it only registers `tags` as an available filter and doesn't call `super`, so `company`/`email`/`first_name`/`last_name`/`job_title`/`phone` never make it into `available_filters`. Redmine's `build_from_params` only turns flat params into filters for keys in that set, so these six get silently dropped and the query runs unfiltered. `tags` and `search`, by contrast, correctly narrowed the results to one contact each, so the mechanism checks out, it's specific to which fields this plugin version registers.

Which plugin version/edition did you verify the filters against? If it's a newer or Pro build where those fields are registered filters, that's worth calling out as a version dependency rather than a blanket claim, since anyone on 4.4.7 Light will see a filter parameter that silently does nothing.

Two ways I'd suggest moving forward:
1. Drop the six non-functional named filters (`company`, `first_name`, `last_name`, `middle_name`, `job_title`, `email`, `phone`) and keep `tags`, `search`, `assigned_to_id`, `author_id`, which do filter server-side here. The field-mapping fix ships cleanly on its own.
2. If you can confirm a specific plugin version where the six filters work, gate the claim on that version and note it in the tool docstring so users on older builds aren't misled.

Happy to re-test against a different plugin version if you point me at one.

jztan (changes requested):
Requesting changes per the comment above: the filter half needs to be corrected or version-gated before merge.

jztan:
One more thing: I merged #223 in the meantime, and this branch now conflicts with develop since both touched `_serialization.py`/`issues.py`. Worth a rebase whenever you get to the filter question above.

mmahmed:
Thanks — that's a useful report. I downloaded the 4.4.7 Light build and read it side by side with the Pro build to work out exactly what differs.

For reference, what I develop and verify against: **Redmine 6.1.1**, `redmine_contacts` **4.4.5, PRO version**.

**Pro against Light.** Light's `ContactQuery#initialize_available_filters` is a single line:

```ruby
def initialize_available_filters
  add_available_filter 'tags', type: :contact_tags, values: ..., order: 12
end
```

The Pro build makes 22 `add_available_filter` calls in that method — including all seven of the fields in question — plus `initialize_author_filter`, `initialize_assignee_filter` and `add_custom_fields_filters`. The file is 100 lines on Light against 256 on Pro, with whole methods absent: `sql_for_is_company_field`, the custom-field columns in `available_columns`. So this is the edition split rather than a change between 4.4.5 and 4.4.7.

On the `super` question: both builds declare `class ContactQuery < Query`, and core's `Query#initialize_available_filters` is an empty `# implemented by sub-classes` (`app/models/query.rb:707`), with `available_filters` doing nothing but calling it and memoising (`:727`). So `super` would contribute nothing on either build, and the Pro one doesn't call it either. The difference is only which `add_available_filter` lines each build ships, which means there is no client-side workaround for it.

**Two additions to your list, both pointing the same way.** `assigned_to_id` and `author_id` don't filter on Light either. Neither is registered there, and the API never reaches the code that would otherwise handle them: `find_contacts`, which does carry an `assigned_to_id` clause and a `job_title` one, is only called from `contacts_notes`, not from `index`. The API index runs `retrieve_crm_query` → `build_from_params` (that helper takes the build-from-params branch unconditionally for `api_request?`, `crm_queries_helper.rb:38`) → `results_scope`, so only `available_filters` and the explicit `:search` option ever reach the query. That is also why `search` narrows on Light: `objects_scope` applies `Contact.live_search` to `options[:search]` directly, outside the filter mechanism.

`cf_N` is in the same position. Pro calls `add_custom_fields_filters(ContactCustomField.where(:is_filter => true))`; Light never calls it. So the note I wrote about a custom field filtering only once its **Used as a filter** setting is on describes the Pro build — on Light the setting makes no difference, because no contact custom field is ever a registered filter.

Net effect: on Light 4.4.7, `tags` and `search` are the only two parameters that narrow `/contacts.json`.

**Two ways I can take this, and I'd rather you pick.** Either a `REDMINE_CRM_EDITION` setting (`light` by default, `pro` opt-in), validated the way `REDMINE_OAUTH_DISCOVERY_AS` already is, where the named filters are sent on `pro` and return an error naming the cause on `light` — so both builds behave correctly and neither hands back a silently unfiltered set. That follows `_is_deals_enabled`, which exists for the same reason, deals being absent from Light. Or, if you'd rather keep this change small, the seven contact fields and `author_id` simply stop being `list` filters and stay `create` attributes — the shape `is_company` already has here, there because the plugin's own `is_company` filter answers with the wrong set for every value, including the explicit `f[]`/`op`/`v` spelling — reachable through `filters` for builds that do register them. Named `list` filters would come down to `tags` and `search` either way, those being the two that hold on both builds.

I lean toward the setting, because it is the difference between a Light caller getting an error and getting a wrong answer, and because a named, typed parameter reads as a guarantee that a silently unfiltered result quietly breaks. The smaller version lands the field-mapping half sooner, though, and I'm happy either way.

One thing I'd rather not fold in: `assigned_to_id` is already on `develop` and predates this change, so I've left it as it is rather than alter behaviour outside this change's scope. Worth its own decision.

For the record, the Pro-side evidence — 54 contacts, comparing the returned id set against ground truth computed from the payload rather than checking only for an empty result:

| filter | HTTP | returned | expected | verdict |
|---|---|---|---|---|
| `first_name` | 200 | 1 | 1 | exact match |
| `last_name` | 200 | 1 | 1 | exact match |
| `company` | 200 | 1 | 1 | exact match |
| `job_title` | 200 | 35 | 35 | exact match |
| `email` | 200 | 1 | 1 | exact match |
| `assigned_to_id` | 200 | 18 | 18 | exact match |
| `author_id` | 200 | 1 | 1 | exact match |
| a custom field with **Used as a filter** on | 200 | 28 | 28 | exact match |
| a custom field with it off | 200 | 54 | 20 | returned everything |

That last row is the same silent-unfiltered behaviour you saw, reached from the other direction — which is what makes the Light case worth stating in the docs rather than left to the caller to discover.

The rebase is pushed: the branch sits on top of `develop` with the #223 and #225 conflicts resolved. The filter change follows once the shape looks right to you.

jztan:
@mmahmed thanks for digging into this, the Pro-side table and the custom-field parallel make the case well. Let's go with option 1, the REDMINE_CRM_EDITION setting. A silently unfiltered result is worse than an explicit error, and it matches the precedent _is_deals_enabled already set.

Go ahead whenever you're ready.

mmahmed:
Done — `REDMINE_CRM_EDITION` is in, pushed as `773ae26`, and CI is green on 3.10 through 3.14. I folded it into this pull request rather than opening a second one, since the filters and their gate aren't much use apart.

The setting takes `light` or `pro` and defaults to `light`, validated on read the way `REDMINE_OAUTH_DISCOVERY_AS` is, so `REDMINE_CRM_EDITION=Professional` is an error rather than a silent fall back to the default. Defaulting to the build that registers fewer filters means an unconfigured deployment refuses a filter it may not be able to apply, instead of answering wrongly.

On `pro`, `first_name`, `last_name`, `middle_name`, `company`, `job_title`, `email`, `phone` and `author_id` go to Redmine as before. On `light` they come back as an error naming every one the call passed — all of them at once, so a caller doesn't discover them one request at a time — and saying what to set and what to do instead:

```
company, job_title: the CRM plugin's Light build does not register these as
query filters, and Redmine ignores an unregistered filter without erroring --
the list would come back unfiltered rather than empty, which is why this is
refused instead of sent. Set REDMINE_CRM_EDITION=pro if this Redmine runs the
Pro build, or filter the returned contacts locally.
```

Three things deliberately left outside the gate. `tags` and `search` work on both builds, so neither is touched. `assigned_to_id` is absent from Light's `ContactQuery` too, but it is already on `develop` and predates this change, so it keeps its current behaviour rather than becoming a new refusal — worth its own decision, and I'd rather not fold it in here. And `filters` is not gated: it never promised that a build honours a given key, so it stays the route that still works on an install which hasn't declared itself.

Documented in `.env.example`, `.env.docker.example`, the README settings table and the plugin section, and `docs/tool-reference.md`. The docstring notes the requirement per parameter, and the tool description says which parameters narrow a list on every build — the grouped `Args` entry the six shared before turned out to reach the schema as one name, leaving five with no description at all, so they're now listed individually.

New tests cover each of the eight refused by default, several reported in one error, an unknown value refused rather than guessed, the value read case- and whitespace-insensitively, and that `tags` / `search` / `assigned_to_id` / `filters` still go through on an undeclared install. 1823 passing locally, same as CI.

Good to merge from my side.

jztan:
Thanks @mmahmed, re-verified live against the sandbox: the eight filters get refused with the combined error on Light, tags/search/assigned_to_id/filters still work, an invalid edition value is rejected, and forcing REDMINE_CRM_EDITION=pro correctly bypasses the guard. All matches what you described. Merging.

Catching assigned_to_id and author_id independently was a good find, and the custom-field parallel made the version-split case clear.

jztan (approved):

PR #229, opened 2026-08-21
feat: return custom fields and relations from the issue list

### LINKED ISSUE #228: `list_redmine_issues` discards the custom fields and relations Redmine already sent
### Problem Statement

Reading a custom field value, or a relation, across a set of issues costs one API request per issue — even though Redmine already returns that data in the list response the server just made and then discards.

`list_redmine_issues` drops it twice over:

- It serializes through `_issue_to_dict_selective`, which delegates to `_issue_to_dict` with `include_custom_fields=False`, so no `custom_fields` key is ever emitted. Naming `custom_fields` in the `fields` parameter does not help either — unknown names are silently skipped.
- The same tool has no `relations` key at all, even though Redmine renders relations on the list endpoint under `include=relations`.

Redmine sends both. `app/views/issues/index.api.rsb` calls `render_api_custom_values issue.visible_custom_field_values, api` unconditionally — no `include` parameter needed — and the same template renders a `relations` array under `include_in_api_response?('relations')`.

So for a 200-issue project, reading one custom field across it is 200 `get_redmine_issue` calls where 2 paged list calls would do. The relations half is what makes a dependency or reference graph readable at all without a per-issue call.

### Proposed Solution

Add opt-in flags that surface the data already in the payload:

- `list_redmine_issues(include_custom_fields=True)` → adds `custom_fields`
- `list_redmine_issues(include_relations=True)` → requests `include=relations` and adds `relations`

Naming `custom_fields` or `relations` in the existing `fields` parameter should do the same thing, and for `relations` that has to add the include to the outgoing request — otherwise the key comes back empty and the caller cannot tell why.

Defaults stay off so existing responses keep their current size, which matters given the documented token budget. The names match the flags `get_redmine_issue` already uses for the same two concepts, and the values keep Redmine's own key names.

No new scope is required: `view_issues` already covers this endpoint, and Redmine has applied per-caller custom field visibility before serializing (`Issue#visible_custom_field_values` intersects a field's roles with the caller's via `visible_by?`).

### Use Case

As an MCP client, I want to read a custom field across every issue in a project in one call so that a routine lookup does not cost hundreds of requests.

Example workflow:
1. `list_redmine_issues(project_id="x", limit=100, include_custom_fields=True)`
2. Read the `custom_fields` array on each returned issue.
3. Page once more for the remainder.

Today step 1 returns no custom fields, so the same result needs one `get_redmine_issue` call per issue.

### Feature Category

Existing Tool Enhancement

### Priority

Important

### Alternatives Considered

- **Filtering server-side with `filters={"cf_N": "value"}`.** Works only when that custom field has "Used as a filter" enabled in Redmine; otherwise `Query#add_filter` discards the filter silently and returns the unfiltered set, which reads like a filter that matched everything. It also does not help when the value itself is what you need to read.
- **A batched multi-issue read tool.** More surface area for less benefit: a serializer key needs no new tool and covers every existing caller.
- **Making the keys unconditional.** Rejected — it would grow every existing list response, and the tool documents a token budget.
- **Reading relations through `getattr(issue, "relations")`.** That is the per-issue request this change exists to remove: python-redmine lists `relations` in `Issue._relations`, and `BaseResource.__getattr__` tests `_relations` before `_includes`, so the attribute ignores an `include=relations` payload and fetches `GET /issues/{id}/relations.json` on first iteration.

### Additional Context

One asymmetry worth recording, since it affects how a consumer should treat the relations half. Redmine filters relations by target-issue visibility on `GET /issues/{id}` — `issues_controller.rb` selects on `other_issue(@issue)&.visible?` — but not on `GET /issues.json`, where the list template iterates `issue.relations` unfiltered. So `issue_to_id` in list output can name an issue the caller cannot read.

That is the endpoint's own behaviour and identical to calling it directly with the same credentials, and filtering client-side would cost one request per relation target, which is the overhead this change exists to remove. But it means a relation edge cannot be classified or counted by its target from list output alone, and that belongs in the docs rather than left for consumers to discover.

The equivalent gap on the projects side — `list_redmine_projects` dropping project custom field values — is filed separately, since it sits in a different file and neither change depends on the other.

### Contribution

mmahmed:
## Summary

Redmine returns issue custom field values on `GET /issues.json` unconditionally, and renders relations there under `include=relations`. `list_redmine_issues` dropped both, so reading a custom field or a relation across a project cost one request per issue.

This adds opt-in flags that surface data already in the payload. Defaults are unchanged, so existing responses keep their current shape and size, and no new scope is required.

Fixes #228

This targets `develop` directly and has no open dependency. It does build on #223, now merged, for the `_included_list` helper it uses to read an include from the response payload — reading relations through `getattr(issue, "relations")` instead would reintroduce exactly the per-issue request that change removed.

## Changes

- `list_redmine_issues` gains `include_custom_fields` and `include_relations`, both defaulting to `False`.
- Naming `custom_fields` or `relations` in `fields` does the same as the matching flag. For `relations` that also adds the include to the outgoing request, so the key cannot come back empty because only one of the two was set. Conversely, setting a flag adds its key to a narrowed `fields` list rather than dropping it, so the flag means the same thing in both forms.
- An `include` supplied through `filters` is normalized and merged rather than overwritten, accepting the string and list forms python-redmine both supports. The pagination count query drops it, since `total_count` comes off the envelope.
- `search_redmine_issues` shares this serializer, so it can now select `custom_fields` too. It hydrates its results through the issues endpoint, which renders the values, so the key was already paid for and discarded. `relations` stays unavailable there: that tool never requests the include, so the key could only ever be empty. Both halves are documented and tested.
- Neither key is built unless it was selected. The serializer computes its field map eagerly, so an unconditional entry would be paid for on every issue of every page — by exactly the callers who narrowed `fields` to avoid that.
- `["*"]` and `["all"]` now behave the same given a tuple, since the request-side guard already accepted any sequence and the sentinel comparison did not.
- Add `_normalize_csv_list` to `_serialization.py` for the include merge. `_normalize_tag_list` delegates to it rather than keeping a second copy of the same coercion, and `docs/contributing.md`'s helper index gains the new name.
- Document the relations visibility asymmetry: Redmine filters relations by target-issue visibility on `GET /issues/{id}` but not on `GET /issues.json`, so a relation edge cannot be classified or counted by its target from list output alone.

Relations are read from the response payload, so a page of issues stays one request and needs only `view_issues`. Custom field values need no include at all — Redmine renders them unconditionally — and arrive already filtered per caller by `Issue#visible_custom_field_values`.

## Testing

`uv run pytest -m "not integration" tests/` — 1777 passed, 0 failed.
`python tests/run_tests.py --all` — same, plus one pre-existing failure:
`test_ssl_integration.py::test_ssl_cert_file_exists` wants a local certificate
file. It fails identically on `develop`, it is `@pytest.mark.integration`, and CI
deselects it.
`uv run black --check src/` and `uv run flake8 src/ --max-line-length=88` — clean.

New `tests/test_list_issue_custom_fields_and_relations.py` (20 tests) covers the new keys, the defaults staying absent, `fields` selection implying the request-side include, a flag adding its key to a narrowed `fields` list, `fields` and the `["*"]` sentinel given as tuples, a caller's own `include` surviving in string, list and tuple form without being duplicated, the count query dropping the include, per-issue values not being shared across a page, and both halves of the shared serializer in `search_redmine_issues` — `custom_fields` selectable, `relations` withheld.

The issue fakes raise `ForbiddenError` on `issue.relations`, so a regression to the lazy attribute fails the suite rather than quietly costing a request per issue.

Run against `develop`'s code, 17 of the 20 fail. The 3 that pass are the regression pins, by design: both `test_absent_by_default` cases and `test_search_does_not_offer_an_empty_relations_key` assert behaviour that must not change, so they hold on either side. Verified separately, with real `redminelib` resources, that neither key costs a request: `custom_fields` is in neither `Issue._includes` nor `Issue._relations`, so it reads from the decoded payload, and relations go through `raw()`. With the engine rigged to raise on any request, a whole page serializes without one.

## Checklist

- [x] `python tests/run_tests.py --all` passes (activate `.venv` first) — apart from the pre-existing `@pytest.mark.integration` SSL-certificate case noted above, which CI deselects
- [x] No manual version bumps
- [x] Docs updated where behavior changed: `docs/tool-reference.md`, `docs/contributing.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   signature serves all seven, and with no `Args` section nothing stated that
   `is_company` writes rather than filters, or that `fields` is for `create` and
   `update` only.
+- `list_redmine_issues` can return `custom_fields` and `relations`, both
+  opt-in. Reading either across a project previously cost one request per
+  issue, because the list serializer dropped data Redmine had already sent:
+  `render_api_custom_values` runs unconditionally on `issues/index.api.rsb`,
+  and the list renders `relations` under `include=relations`. Pass
+  `include_custom_fields=True` / `include_relations=True`, or name
+  `custom_fields` / `relations` in `fields` -- naming `relations` there also
+  adds the include to the request, so the key can never come back
+  misleadingly empty, and setting a flag adds its key to a narrowed `fields`
+  list rather than dropping it. Relations are read from the response payload,
+  so a page of issues stays one request and needs only `view_issues`. An
+  `include` passed through `filters` is preserved and merged, given as either
+  a string or a list. Defaults are unchanged, so existing responses keep their
+  current size, and no new scope is required.
+  ([#228](https://github.com/jztan/redmine-mcp-server/issues/228))
+- `search_redmine_issues` can select `custom_fields` too, as a consequence of
+  sharing the issue serializer: it hydrates its results through the issues
+  endpoint, which renders the values, so the key was already paid for and
+  discarded. `relations` is deliberately not offered there, since that tool
+  never requests the include and the key could only ever be empty.
+
+  One asymmetry: Redmine filters custom field values per caller
+  (`Issue#visible_custom_field_values`), but filters relations by target-issue
+  visibility only on `GET /issues/{id}`, not on `GET /issues.json`. So
+  `issue_to_id` in list output can name an issue the caller cannot read. This
+  exposes nothing a caller could not get from the same endpoint directly, and
+  filtering it client-side would cost one request per relation target -- the
+  overhead this option exists to remove. Resolve a target through
+  `get_redmine_issue`, which does apply the filter.
 
 ### Fixed
 - `include=relations` is no longer discarded and re-fetched per issue.
@@ -168,6 +197,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   [#227](https://github.com/jztan/redmine-mcp-server/pull/227))
 
 ### Tests
+- New `tests/test_list_issue_custom_fields_and_relations.py` covers the new
+  list keys, the defaults staying absent, `fields` selection implying the
+  request-side include, a caller's own `include` surviving in string, list and
+  tuple form without being duplicated, a flag adding its key to a narrowed
+  `fields` list, the `["*"]` sentinel behaving the same as a tuple, the
+  pagination count query dropping the include, per-issue values not being
+  shared across a page, and both halves of the shared serializer in
+  `search_redmine_issues` -- `custom_fields` selectable, `relations` withheld.
+  The issue fakes raise on `issue.relations`, so a regression to the lazy
+  attribute fails rather than quietly slowing down.
 - New `tests/test_relations_payload.py` covers reading relations from the
   include payload, an absent include, an empty include, the Gantt path issuing
   one request for a whole chart, the leaf-issue `children` case, and the delete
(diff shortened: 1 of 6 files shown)


jztan:
Thanks @mmahmed, re-verified live against the sandbox: default response shape unchanged, include_custom_fields returns real values, include_relations returned a relation I created and then went back to empty after I deleted it, and search_redmine_issues withholds relations while allowing custom_fields, exactly as documented. Merging.

The relations-visibility note (list endpoint not filtering by target-issue visibility, unlike the single-issue endpoint) was a sharp catch and good to have documented rather than silently exposed.

jztan (approved):

PR #231, opened 2026-08-21
feat: return project custom field values from the project list

### LINKED ISSUE #230: `list_redmine_projects` discards the project custom field values Redmine already sent
### Problem Statement

Nothing in the tool surface returns project custom field values. Not `list_redmine_projects`, not `get_project_dashboard_data`, not `get_project_modules`, not issue payloads (which carry only `project: {id, name}`), and `search_entire_redmine` does not accept projects as a scope.

Redmine does return them. `app/views/projects/index.api.rsb` calls `render_api_custom_values project.visible_custom_field_values, api` unconditionally — no `include` parameter needed. `list_redmine_projects` then builds a five-key dict literal (`id`, `name`, `identifier`, `description`, `created_on`) and drops everything else in the payload.

So a value the server already fetched is unreachable, and discovering which projects carry a given custom field value costs one call per project.

There is also an asymmetry that makes this look like an oversight rather than a design choice: `list_project_issue_custom_fields` returns the *definitions* of a project's issue custom fields, so the tool surface describes the fields a project has without ever returning the project's own values for them.

### Proposed Solution

`list_redmine_projects(include_custom_fields=True)` adds a `custom_fields` array to each project, read from the response payload.

Default off, so existing responses keep their current size — this matters given the documented token budget. The parameter name matches the flag `get_redmine_issue` and `list_redmine_issues` already use for the same concept, and the values keep Redmine's own key names.

No new scope is required: `view_project` already covers this endpoint, and Redmine has applied per-caller custom field visibility before serializing (`Project#visible_custom_field_values` intersects a field's roles with the caller's via `visible_by?`).

### Use Case

As an MCP client, I want to find the projects whose owner custom field names a particular person, in one call rather than one per project.

Example workflow:
1. `list_redmine_projects(include_custom_fields=True)`
2. Select the projects whose `custom_fields` entry matches.

Today step 1 returns no custom fields, and there is no other tool that would, so the only route is one call per project — with the qualifying attribute being exactly the one that is missing from the list output.

### Feature Category

Existing Tool Enhancement

### Priority

Important

### Alternatives Considered

- **Adding a single-project getter** (there is none today). Useful on its own, but it does not solve discovery: finding which projects qualify still costs one call per project, because the qualifying attribute is the field being fetched.
- **Filtering issues on a project custom field with `filters={"cf_N": "value"}`.** Does not work — a project custom field is not in an issue query's `available_filters`, so `Query#add_filter` discards it silently and returns the unfiltered set.
- **Making the key unconditional.** Rejected — it would grow every existing `list_redmine_projects` response, and the tool documents a token budget.
- **Reading the values through `list_project_issue_custom_fields`.** Returns definitions, not values, and only for issue custom fields.

### Additional Context

The equivalent gap on the issues side — `list_redmine_issues` dropping issue custom field values and relations — is [#228](https://github.com/jztan/redmine-mcp-server/issues/228), filed separately since it sits in a different file and neither change depends on the other.

### Contribution

mmahmed:
## Summary

Redmine renders project custom field values on `GET /projects.json` unconditionally, but `list_redmine_projects` built a fixed five-key dict and dropped them. No other tool returns them either — there is no single-project getter — so a value the server had already fetched was not reachable through this server at all.

This adds one opt-in flag that surfaces data already in the payload. The default output is unchanged and no new scope is required.

Fixes #230

This reuses the `_custom_fields_to_list` helper #227 moved into `_serialization.py`, rather than adding a third copy of the same coercion.

## Changes

- `list_redmine_projects` gains `include_custom_fields`, defaulting to `False`. When set, the values come back with the projects and no per-project follow-up request is made.
- The docstring distinguishes project custom fields from issue custom fields — separate things in Redmine, with only the latter covered by `list_project_issue_custom_fields` and no tool exposing project custom field definitions. It sits above `Args:`, which is the part of a docstring that reaches a calling model.

No new scope: `view_project` is enough, and Redmine has already limited the values to those the caller may see via `Project#visible_custom_field_values`.

## Testing

`uv run pytest -m "not integration" tests/` — 1827 passed, which is what CI runs. `python tests/run_tests.py --all` leaves one failure, `test_ssl_integration.py::test_ssl_cert_file_exists`, which fails identically on `develop` and is deselected by CI.
`uv run black --check src/` and `uv run flake8 src/ tests/ --max-line-length=88` — clean.

New `tests/test_project_custom_field_values.py` (4 tests) covers the new key, the five existing keys staying exactly as they are with the flag off, a project whose payload omits `custom_fields` entirely — what Redmine sends when there are no visible values — and distinct values landing on their own projects with no per-project fetch.

The fakes expose custom fields as attributes rather than plain dicts, because that is what python-redmine hands the serializer. A fixture built from dicts would pass even if the serializer were bypassed; these fail if it is.

Three of the four fail against the parent commit; the fourth is a regression pin on the existing five-key output, so it passes there by design. Verified separately, with real `redminelib` resources, that `custom_fields` is in neither `Project._includes` nor `Project._relations`, so reading it issues no HTTP request.

## Checklist

- [x] `uv run pytest -m "not integration" tests/` passes — 1827. The one `run_tests.py --all` failure is a pre-existing integration case, unchanged from `develop`
- [x] `uv run black --check src/` and `uv run flake8 src/ tests/ --max-line-length=88` are clean
- [x] No manual version bumps
- [x] Docs updated where behavior changed: `docs/tool-reference.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `is_company` stays create-only: the plugin's own filter answers with the same
   wrong set for every value, including the explicit `f[]`/`op`/`v` spelling.
 
+- `list_redmine_projects` can return `custom_fields`, opt-in. Redmine renders
+  project custom field values on `GET /projects.json` unconditionally
+  (`render_api_custom_values` in `projects/index.api.rsb`), but the tool built
+  a fixed five-key dict and dropped them -- and no other tool returned them
+  either, so project custom field values were not reachable through this
+  server at all. Pass `include_custom_fields=True` and they come back with the
+  projects, at no additional request per project. The default output is
+  unchanged and no new scope is required -- Redmine has already limited the
+  values to those the caller may see, and `view_project` is enough
+  ([#230](https://github.com/jztan/redmine-mcp-server/issues/230)).
 - `manage_contact` can filter a `list` server-side on `first_name`, `last_name`,
   `middle_name`, `company`, `job_title`, `email`, `phone` and `author_id`. Seven
   of those already existed on the tool as create-only parameters and were
@@ -178,6 +188,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   exposed `relations` as a plain attribute, which no python-redmine issue ever
   does; they now serve includes from `raw()` and raise on the attribute, so
   this class of bug fails in CI instead of only against a live Redmine.
+- New `tests/test_project_custom_field_values.py` covers the new key, the five
+  existing keys staying exactly as they were, a project whose payload omits
+  `custom_fields` entirely (what Redmine sends when there are none), values
+  arriving as resource objects rather than plain dicts, and no per-project
+  follow-up request.
 - New `tests/test_contact_payload_keys.py` covers the contact serializer against
   the shape the CRM API actually returns -- emails as an array, tags under
   `tag_list`, the plugin's own address field names, and the `custom_fields`,
--- a/docs/tool-reference.md
+++ b/docs/tool-reference.md
@@ -210,9 +210,10 @@ This allows LLM consumers to distinguish trusted tool output from untrusted user
 
 Lists all accessible projects in the Redmine instance.
 
-**Parameters:** None
+**Parameters:**
+- `include_custom_fields` (boolean, optional): Add `custom_fields` to each project. Default: `false`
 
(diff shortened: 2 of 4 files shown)


jztan:
Thanks @mmahmed, verified live against the sandbox: default response shape unchanged, include_custom_fields returns [] when no project custom field exists, and I created a real project custom field (Team Size = Medium) via the Rails console to confirm the non-empty path serializes correctly ({id, name, value}), then cleaned it up. Merging.

jztan (approved):

Issue #119, opened 2026-05-15
list_project_issue_custom_fields reports is_required=false for fields that create_redmine_issue requires


jztan:
## Problem

When \`list_project_issue_custom_fields\` returns \`is_required: false\` for a custom field, an agent reasonably concludes the field can be omitted from \`create_redmine_issue\`. In practice, the create call rejects with \`\"Validation failed: <field> cannot be blank\"\`, costing the agent at least one wasted turn -- the exact thing discovery tools are supposed to prevent.

## Reproduction

1. \`list_project_issue_custom_fields(project_id=1)\` returns:
   \`\`\`json
   [
     {\"id\": 1, \"name\": \"Priority Level\", \"field_format\": \"\", \"is_required\": false, ...},
     {\"id\": 2, \"name\": \"Department\",     \"field_format\": \"\", \"is_required\": false, ...}
   ]
   \`\`\`
2. \`create_redmine_issue(project_id=1, subject=\"...\")\` (no custom fields supplied) returns:
   \`\`\`json
   {\"error\": \"Validation failed: Department cannot be blank\"}
   \`\`\`
3. Workaround that succeeded: supplying \`extra_fields={\"custom_fields\": [{\"id\": 2, \"value\": \"Engineering\"}]}\` on the create call.

**Confirmed not a tracker-scope issue:** re-ran step 1 with \`tracker_id=1\` (bug tracker) and Department still reports \`is_required: false\`. Passing the tracker scope is not the fix; the discovery tool genuinely doesn't reflect the runtime requirement.

The \`trackers: []\` array being empty on both custom fields is also suspicious -- if these fields are scoped to specific trackers in Redmine, that array should be populated. Possibly a related metadata-fetch issue, or the requirement comes from elsewhere entirely (workflow rule, role binding, project-level setting). Worth investigating alongside the main bug.

## Likely cause (needs confirmation)

Redmine's \`is_required\` flag on custom field definitions is set at the field-definition level, but enforcement can also come from tracker-level workflow rules or from the field being marked required for a specific role/tracker combination. \`GET /custom_fields.json\` (the underlying call) doesn't reflect these contextual rules. The Redmine UI computes the \"effective required\" status by intersecting field def + tracker config + role + workflow state, which the bare API doesn't expose.

## Suggested fix (in order of preference)

[...]

PR #233, opened 2026-08-24
fix: stop inventing custom field metadata Redmine never sent

### LINKED ISSUE #232: `list_project_issue_custom_fields` invents six metadata keys Redmine never sends
## Bug Description

`list_project_issue_custom_fields` returns eight keys per field. Redmine sends two. The other six are `getattr` fallbacks, so the tool reports `field_format: ""`, `is_required: false`, `multiple: false`, `default_value: null`, `possible_values: []` and `trackers: []` for every custom field, always, on every deployment.

`is_required: false` is the sharp edge. It is byte-identical to a genuine "this field is optional", so a caller cannot detect that it was invented. It acts on the value, omits the field, and Redmine rejects the create with `"<field> cannot be blank"` — which is exactly the failure reported in #119.

**#119 is this bug, closed on a diagnosis that does not hold.** The probe in that thread created an issue supplying only `Department`, found `Priority Level` was genuinely optional, and concluded "not a systemic metadata-fetch failure (Priority Level is classified correctly)". Priority Level was not classified. Both fields returned the same hardcoded `false`; one of them happened to be right. A constant `false` matches every genuinely-optional field in every project, so that probe cannot distinguish "correct classification" from "constant that coincides". The contextual-rules caveat #119 landed is real and worth keeping, but it is not why `is_required` is always false. #119 is locked, hence a new issue.

Absence would be safe here. A caller can detect `null` and ask a human. Fabrication is not detectable, and that is the whole of this report.

## Steps to Reproduce

1. Point the server at any Redmine (6.1, 7.0 and trunk all behave identically here) with a project that has issue custom fields, at least one of them a **list** field with possible values and at least one marked **Required**.
2. Call `list_project_issue_custom_fields(project_id=<id>)`.
3. Every entry comes back with `field_format: ""`, `is_required: false`, `multiple: false`, `default_value: null`, `possible_values: []`, `trackers: []`, regardless of how the fields are actually configured.
4. Call `create_redmine_issue` for that project without supplying the required field. It is rejected with `"<field> cannot be blank"` for a field step 3 reported as not required.

Confirmed against a live 6.1 instance with a non-admin API key: 18 fields returned, every one with `field_format: ""` and `possible_values: []`, including list fields with known values, and the field marked Required reported as `is_required: false`.

## Expected Behavior

Only what Redmine actually sent. `id` and `name` carry data; the other six keys report `null`, documented as "not readable on this token" — never a plausible-looking default like `false` or `[]`.

Acceptance test: on a custom field marked **Required** in Redmine's admin UI, `is_required` must not come back `false`. Today it does.

## Actual Behavior

Six of the eight keys are invented locally and are indistinguishable from real answers.

The tool calls `project.get(project_id, include="issue_custom_fields")` (`tools/projects.py`). Redmine renders that include as exactly `id` and `name`:

```ruby
# app/helpers/projects_helper.rb
api.array :issue_custom_fields do
  project.all_issue_custom_fields.each do |custom_field|
    api.custom_field(:id => custom_field.id, :name => custom_field.name)
  end
end if include_in_api_response?('issue_custom_fields') && User.current.allowed_to?(:view_issues, project)
```

Byte-identical on `6.1-stable`, `7.0-stable` and `master`. So `_custom_field_to_dict` reads six attributes that are not in the payload. python-redmine raises `ResourceAttrError` for each, which subclasses `AttributeError`, so the three-argument `getattr` swallows it and substitutes the default. `CustomField._relations` is empty, so nothing is lazily fetched either — the values are pure local invention.

The real definitions live on `GET /custom_fields.json`, which Redmine restricts to administrators (`before_action :require_admin`, covering `index` in 6.1 and 7.0; `accept_api_auth :index` permits API *auth*, not non-admins). There is no permission and no setting that relaxes it — `require_admin` is a bare `User.current.admin?` and no custom-field permission is registered in `lib/redmine.rb`. Under OAuth it is stricter still: `User#admin?` is `super and @oauth_scope.include?(:admin)` (`app/models/user.rb`), so even a Redmine administrator's token gets 403 unless it carries the `admin` scope — which this server deliberately never advertises, and rightly so.

**This is not a gap Redmine is about to close.** Four open requests ask for exactly this, the oldest from 2015, none with a target version:

- [#18875](https://www.redmine.org/issues/18875) — why does `GET /custom_fields` require admin (Defect, New)
- [#41318](https://www.redmine.org/issues/41318) — REST API "Custom Fields" needs administration privileges? (Feature, New)
- [#42581](https://www.redmine.org/issues/42581) — add an endpoint for `possible_values` (Feature, New)
- [#43407](https://www.redmine.org/issues/43407) — add more metadata to custom field definitions in the REST API (Feature, New), which asks for `field_format`, `multiple`, `is_required`, `default_value` and `possible_values` on this exact include

[#33301](https://www.redmine.org/issues/33301), the feature that introduced the include in 4.2.0, deliberately shipped id and name only. So there is no endpoint, permission, scope or configuration that makes this data available to a normal token. The only thing in this server's control is whether it invents the values or reports that it does not have them.

Three things follow from the same root cause and are worth fixing together:

**`tracker_id` cannot ever filter.** `_custom_field_applies_to_tracker` treats empty tracker bindings as "globally available", and bindings are always empty, so the parameter silently returns every field in the project. A caller that asked for a filtered list gets an unfiltered one and cannot tell. #119 re-ran the call with `tracker_id=1`, got the same answer, and read that as evidence about required-ness rather than about the filter. (Separately, "empty means global" is backwards: `Issue#available_custom_fields` is `project.all_issue_custom_fields & tracker.custom_fields`, so a field bound to no tracker reaches no issue.)

**Autofill's `default_value` path is dead.** `REDMINE_AUTOFILL_REQUIRED_CUSTOM_FIELDS` resolves a value from `custom_field.default_value` or the env map. `default_value` is never in the payload, so only `REDMINE_REQUIRED_CUSTOM_FIELD_DEFAULTS` can ever supply one. `_is_allowed_custom_field_value` is likewise inert, since `possible_values` is always empty and it returns True for anything.

**The scope map is justified by the wrong endpoint, and is short two permissions.** `TOOL_SCOPES` maps this tool to `frozenset()` with the comment "GET /custom_fields.json is admin-gated by Redmine itself". The tool has never called that endpoint — `git log -S 'custom_fields.json' -- src` finds it only in comments and docstrings, both added by the #119 fix. The include it does call is rendered only for a caller holding `view_issues`, and is omitted otherwise, so a token without `view_issues` gets `[]`, which reads as "this project has no custom fields".

`view_project` is missing too, and that one produces a worse failure than a wrong `[]`. The tool calls `projects#show`, which is gated on `view_project`. That permission is declared `:public => true`, which looks like an exemption and is not one — `Role#allowed_permissions` builds `unscoped = permissions + Redmine::AccessControl.public_permissions` and then returns `unscoped & scope`, so the public permissions are intersected with the token's scopes as well. A token scoped only `view_issues` is therefore admitted by `ScopeEnforcementMiddleware`, which believes the tool's declared requirements are met, and is then refused by Redmine with an opaque 403 — the outcome the scope map exists to turn into a clear `INSUFFICIENT_SCOPE` denial. `summarize_project_status` already maps both for the same `project.get()` call.

Two documented claims are also unachievable. The docstring names `GET /custom_fields.json` as "the underlying API", and `docs/tool-reference.md` advertises the tool as returning "allowed values and tracker bindings" with an example output showing `field_format: "list"`, `possible_values: ["S", "M", "L"]` and a populated `trackers` array. No token can produce that response.

Why the test suite is green: every fixture builds the field with `unittest.mock.Mock` and sets all eight attributes by hand. A `Mock` answers any attribute access, so a serializer reading attributes Redmine never sends looks correct under it. The suite validates a payload shape that does not exist.

## Version

2.12.0

## Deployment Method

Docker

## MCP Client

Claude Code (CLI)

## Redmine Version

6.1

## Additional Context

A pull request follows immediately: `null` for the six unpopulatable keys with the keys kept for schema stability, an explicit error when `tracker_id` cannot be honoured, `view_project` and `view_issues` on the scope map, the docstring and `docs/tool-reference.md` corrected, and a test module that uses fakes raising `ResourceAttrError` instead of `Mock` so absence is actually exercised.

For the record on one alternative considered and rejected: trying `GET /custom_fields.json` first and falling back to the include would give admin-key deployments real metadata, but it cannot work at all in either OAuth mode without advertising the `admin` scope, it costs every other deployment a guaranteed 403 per call, and it makes the response shape depend on the caller's privileges. Reporting `null` honestly is the smaller and more predictable fix.

mmahmed:
## Summary

Fixes #232. `list_project_issue_custom_fields` advertised eight keys per custom field and Redmine only sends two, so six of them were `getattr` fallbacks: `field_format: ""`, `is_required: false`, `multiple: false`, `default_value: null`, `possible_values: []`, `trackers: []`, for every field, always. Those six now report `null`, documented as "not readable on this token".

`is_required: false` was the one that cost callers turns. It is byte-identical to a real "this field is optional", so nothing downstream could detect it was invented — the caller omitted the field and Redmine rejected the create with `"<field> cannot be blank"`. That is the failure #119 reported. #119 landed a docstring caveat about workflow rules and role-based field permissions, which is accurate and is kept here, but it is not why the flag is always false: the flag is a local default that never came from Redmine at all.

## Why null rather than the admin endpoint

The definitions live on `GET /custom_fields.json`, which Redmine restricts to administrators (`before_action :require_admin`, still covering `index` in 7.0). No permission or setting relaxes it: `require_admin` is a bare `User.current.admin?` and no custom-field permission is registered in `lib/redmine.rb`. Under OAuth it is stricter — `User#admin?` is `super and @oauth_scope.include?(:admin)`, so even an administrator's token is refused unless it carries `admin`, which this server never advertises.

So a fallback to that endpoint would help only API-key deployments whose key belongs to an admin, cost everyone else a guaranteed 403 per call, and make the response shape depend on the caller's privileges. Reporting absence honestly is the smaller and more predictable fix.

Redmine is not about to close the gap from its side either: [#18875](https://www.redmine.org/issues/18875), [#41318](https://www.redmine.org/issues/41318), [#42581](https://www.redmine.org/issues/42581) and [#43407](https://www.redmine.org/issues/43407) all ask for this metadata on a non-admin path, the oldest from 2015, none with a target version. #43407 asks for exactly these five keys on exactly this include.

The keys are kept rather than dropped, so the response shape stays stable across deployments and so they start carrying real data for free if a future Redmine populates the include.

## Changes

Each section stands alone; drop any hunk you disagree with.

**`_custom_field_to_dict` reports absence as absence.** Every key is read through a new `_payload_attr` helper in `_serialization.py`, which takes the coercion as an argument and applies it *only* to a value the payload actually carried. That is the part that matters: `bool` and the `possible_values` normalizer can no longer run on a key Redmine never sent, so there is no spelling of this serializer that turns an absent key back into `false` or `[]`. python-redmine raises `ResourceAttrError`, an `AttributeError` subclass, for an attribute absent from a response, which is what a three-argument `getattr` was silently swallowing; a client configured with `raise_attr_exception=False` returns `None` instead, and both land as `null`. Real data still passes through untouched, including `possible_values` in the `[{"value": ...}]` shape the admin endpoint uses, so the serializer is correct for any payload rather than hardcoded for one.

It lives in `_serialization.py` rather than in `tools/projects.py` because the same "an absent key became a plausible default" pattern is present in several sibling serializers; this puts the primitive where those fixes can reuse it instead of each inventing its own sentinel. `_extract_possible_values` gains a `_normalize_possible_values` half so the attribute is read once, under one absence rule, rather than twice under two.

`name` also loses its `""` default. Redmine always sends it, so the default was unreachable, but `""` is the same category of invented answer.

**`tracker_id` errors instead of returning an unfiltered list.** Tracker bindings are part of the admin-only definition, so `_custom_field_applies_to_tracker` was reading absent bindings as "applies to every tracker" and the filter silently returned every field in the project — a caller could not tell a filtered list from an unfiltered one. Passing `tracker_id` when bindings are unreadable now returns an `error` with `code: "TRACKER_BINDINGS_UNREADABLE"` and the recovery in `hint`, matching the shape the existing refused preconditions use (`CONFIRMATION_REQUIRED`, `CHILDREN_PRESENT`, `ATTACHMENT_UNAVAILABLE`). The `error` describes what *this* response carried rather than asserting what Redmine always does, since bindings are readable on a payload that includes them and a fixed explanation would be wrong for that deployment; the standing Redmine reason sits in the `hint`. An empty project is not an error: there is nothing to filter.

Where bindings *are* readable, a field bound to no tracker is now excluded rather than treated as global. `Issue#available_custom_fields` is `project.all_issue_custom_fields & tracker.custom_fields`, so an unbound field intersects to nothing and reaches no issue; "empty means global" had it backwards.

**`view_project` and `view_issues` on the scope map.** `TOOL_SCOPES` mapped this tool to `frozenset()`, justified by a comment about `/custom_fields.json` being admin-gated — an endpoint the tool has never called (`git log -S 'custom_fields.json' -- src` finds it only in comments and docstrings). The include it does call is rendered only for a caller holding `view_issues` and is omitted otherwise, so a token without that scope received `[]`, indistinguishable from "this project has no custom fields". Same reasoning as the existing `list_redmine_versions` entry.

`view_project` is needed as well, for `projects#show` itself. It is declared `:public => true`, which is easy to read as an exemption and is not one: `Role#allowed_permissions` is `unscoped = permissions + Redmine::AccessControl.public_permissions...` followed by `scope.present? ? unscoped & scope : unscoped`, so the public permissions are intersected with the token's scopes too. A token scoped only `view_issues` would therefore be admitted by this server's own middleware and then refused by Redmine with an opaque 403 — the failure the scope map exists to turn into a clear denial. `summarize_project_status` already maps both for the same `project.get()` call.

**Docs.** The docstring no longer names `GET /custom_fields.json` as the underlying API, and the `null` contract moves into the leading prose. That last part is not cosmetic: FastMCP keeps only the leading prose as the tool description and lifts `Args:` entries into the per-parameter schema, dropping everything after them — so the entire "null does not mean false" guarantee was invisible to a model calling the tool, and the surviving one-line description still promised "custom fields configured for a project". Dumping the registered tool confirms both the old omission and the fix. `docs/tool-reference.md` loses the "including allowed values and tracker bindings" headline and the example output that no token can produce, and gains the `null` contract plus links to the four Redmine tracker items. The autofill notes in `README.md` and `docs/tool-reference.md` record that the `default_value` half is inert against a stock Redmine, since `default_value` is not in the include the server reads — only `REDMINE_REQUIRED_CUSTOM_FIELD_DEFAULTS` can supply a value. No behaviour change there; it just no longer promises something it cannot do.

## Tests

New `tests/test_custom_field_metadata_absence.py`, which deliberately does not use `Mock`. That is the point: a `Mock` answers every attribute access, so a serializer reading attributes Redmine never sent looks correct under it, which is why the existing suite was green against the fabricated defaults. The fakes here raise `ResourceAttrError` for absent attributes exactly as python-redmine does, with a second fake covering the `raise_attr_exception=False` client that returns `None` instead.

Covers: the include payload yielding `null` for all six keys, parametrized across both client configurations; the acceptance test from the issue, that a required field does not report `false`; real metadata passing through when a payload does carry it; a genuinely empty binding list staying `[]` rather than becoming `null`; `tracker_id` erroring with its `code` and `hint` when bindings are unreadable, filtering when they are, excluding an unbound field, and not erroring on an empty project.

Each behaviour change was mutation-checked: reverting the serializer to its fabricating defaults, and restoring "no bindings means global", each turn the new tests red. A test that cannot fail is not evidence.

Four existing assertions in `tests/test_custom_field_helpers.py` and one in `tests/test_list_project_issue_custom_fields.py` asserted the old contract (`[]` for absent bindings, empty bindings applying to every tracker). Updated with the reasoning inline rather than deleted.

`1857 passed, 0 failed` on the full non-integration suite, measured from a clean `git archive` of the head rather than a working tree, against a `1847 passed` baseline on `develop`. `flake8` and `black --check` are clean on every file this branch touches.

## Compatibility

Breaking for a caller that reads the six keys as booleans or lists, which is the intent — those values were never Redmine's. A caller doing `if field["is_required"]:` is unaffected, since `null` and `false` are both falsy; one doing `field["is_required"] is False` or `for v in field["possible_values"]` needs a null check. `id` and `name` are unchanged, so name-keyed and id-keyed custom field writes are unaffected, as is `docs/troubleshooting.md`'s advice to find a field id with this tool.

Two behaviour changes beyond the null shape: `tracker_id` now errors where it used to return everything, and the tool now requires the `view_project` and `view_issues` scopes where it required none. Both scopes are ones this server already advertises, so no deployment needs a new consent — a token that could usefully call this tool already carries them, and one that could not now gets a clear denial instead of a 403.

## Verification

Redmine's renderer was read at `6.1-stable`, `7.0-stable` and `master` — `app/helpers/projects_helper.rb` is byte-identical in all three, and the whole API surface was grepped for the definition keys (`field_format|is_required|possible_values|is_filter` across every `*.api.rsb` and helper): only `custom_fields/index.api.rsb` has them. `/trackers.json` exposes `enabled_standard_fields`, core fields only. The one partial exception is `multiple: true` on issue *value* payloads (`custom_fields_helper.rb`), emitted only when true, which is not a definition endpoint and is not used here.

Behaviour was confirmed against a live 6.1 instance with a non-admin API key: the include returns 18 fields whose keys are exactly `id` and `name`, `GET /custom_fields.json` returns 403, and the field marked Required in admin reported `is_required: false` before this change and `null` after.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,7 +51,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   [#42581](https://www.redmine.org/issues/42581) and
   [#43407](https://www.redmine.org/issues/43407), the last asking for exactly
   these keys.
-  ([#ISSUE](https://github.com/jztan/redmine-mcp-server/issues/ISSUE))
+  ([#232](https://github.com/jztan/redmine-mcp-server/issues/232))
 - `list_project_issue_custom_fields` now returns an error --
   `code: "TRACKER_BINDINGS_UNREADABLE"`, with the recovery in `hint` -- when
   `tracker_id` is passed and the response does not describe tracker bindings,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,49 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   path for the `me` id, not `current`. The distinction matters because
   `/my/account.json` renders no memberships at all.
 
+- `list_project_issue_custom_fields` no longer invents the six keys Redmine
+  never sends. It reads
+  `GET /projects/{id}.json?include=issue_custom_fields`, which Redmine renders
+  as exactly `id` and `name` per field
+  (`app/helpers/projects_helper.rb`, unchanged in 6.1, 7.0 and trunk), and
+  every other attribute was read through a `getattr` fallback -- so
+  `field_format: ""`, `is_required: false`, `multiple: false`,
+  `default_value: null`, `possible_values: []` and `trackers: []` came back
+  for every field, always. `is_required: false` was the sharp edge: it is
+  byte-identical to a real "this field is optional", so a caller omitted the
+  field and Redmine rejected the create with `"<field> cannot be blank"` --
+  the failure #119 reported and could not explain. Those keys are now `null`,
+  documented as "not readable on this token". They are kept rather than
+  dropped so the response shape is stable and so they populate for free if a
+  future Redmine fills the include in.
+
+  The definitions live on `GET /custom_fields.json`, which Redmine restricts
+  to administrators (`before_action :require_admin`, still covering `index` in
+  7.0), so no non-admin token can obtain them; under OAuth an administrator's
+  token cannot either, since `User#admin?` also requires the `admin` scope,
+  which this server deliberately never requests. Redmine has four open
+  requests to expose this metadata to non-admins, the oldest from 2015 and
+  none with a target version:
+  [#18875](https://www.redmine.org/issues/18875),
+  [#41318](https://www.redmine.org/issues/41318),
+  [#42581](https://www.redmine.org/issues/42581) and
+  [#43407](https://www.redmine.org/issues/43407), the last asking for exactly
+  these keys.
+  ([#ISSUE](https://github.com/jztan/redmine-mcp-server/issues/ISSUE))
+- `list_project_issue_custom_fields` now returns an error when `tracker_id` is
+  passed and the response does not describe tracker bindings, which on a stock
+  Redmine is always. It used to read absent bindings as "applies to every
+  tracker" and return every field in the project, indistinguishable from a
+  filtered list. Where bindings *are* readable, a field bound to no tracker is
+  now excluded rather than treated as global: Redmine computes an issue's
+  fields as `project.all_issue_custom_fields & tracker.custom_fields`
+  (`app/models/issue.rb`), so an unbound field reaches no issue.
+- `list_project_issue_custom_fields` requires the `view_issues` scope instead
+  of none. Redmine renders the `issue_custom_fields` include only for a caller
+  holding `view_issues`, and omits the array otherwise, so a token without it
+  got `[]` -- which reads as "this project has no custom fields". The scope
+  map justified the empty entry with a comment about `/custom_fields.json`
+  being admin-gated, an endpoint this tool has never called.
 ### Added
 - `list_redmine_projects` now returns the six fields Redmine's project index
   renders and this serializer discarded: `homepage`, `parent`, `status`,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,21 +52,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   [#43407](https://www.redmine.org/issues/43407), the last asking for exactly
   these keys.
   ([#ISSUE](https://github.com/jztan/redmine-mcp-server/issues/ISSUE))
-- `list_project_issue_custom_fields` now returns an error when `tracker_id` is
-  passed and the response does not describe tracker bindings, which on a stock
-  Redmine is always. It used to read absent bindings as "applies to every
+- `list_project_issue_custom_fields` now returns an error --
+  `code: "TRACKER_BINDINGS_UNREADABLE"`, with the recovery in `hint` -- when
+  `tracker_id` is passed and the response does not describe tracker bindings,
+  which on a stock Redmine is always. It used to read absent bindings as "applies to every
   tracker" and return every field in the project, indistinguishable from a
   filtered list. Where bindings *are* readable, a field bound to no tracker is
   now excluded rather than treated as global: Redmine computes an issue's
   fields as `project.all_issue_custom_fields & tracker.custom_fields`
   (`app/models/issue.rb`), so an unbound field reaches no issue.
-- `list_project_issue_custom_fields` requires the `view_issues` scope instead
-  of none. Redmine renders the `issue_custom_fields` include only for a caller
-  holding `view_issues`, and omits the array otherwise, so a token without it
-  got `[]` -- which reads as "this project has no custom fields". The scope
-  map justified the empty entry with a comment about `/custom_fields.json`
-  being admin-gated, an endpoint this tool has never called.
-### Added
+- `list_project_issue_custom_fields` requires the `view_project` and
+  `view_issues` scopes instead of none. Redmine renders the
+  `issue_custom_fields` include only for a caller holding `view_issues`, and
+  omits the array otherwise, so a token without it got `[]` -- which reads as
+  "this project has no custom fields". `view_project` is needed for
+  `projects#show` itself: it is a `:public => true` permission, which is not an
+  exemption, because `Role#allowed_permissions` intersects the public
+  permissions with the token's scopes too (`scope.present? ? unscoped & scope
+  : unscoped`), so a token scoped only `view_issues` is refused by Redmine
+  after this server had already admitted the call. The scope map justified the
+  empty entry with a comment about `/custom_fields.json` being admin-gated, an
+  endpoint this tool has never called.
 - `list_redmine_projects` now returns the six fields Redmine's project index
   renders and this serializer discarded: `homepage`, `parent`, `status`,
   `is_public`, `inherit_members` and `updated_on`. `status` is Redmine's
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,18 +61,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   now excluded rather than treated as global: Redmine computes an issue's
   fields as `project.all_issue_custom_fields & tracker.custom_fields`
   (`app/models/issue.rb`), so an unbound field reaches no issue.
-- `list_project_issue_custom_fields` requires the `view_project` and
-  `view_issues` scopes instead of none. Redmine renders the
-  `issue_custom_fields` include only for a caller holding `view_issues`, and
-  omits the array otherwise, so a token without it got `[]` -- which reads as
-  "this project has no custom fields". `view_project` is needed for
-  `projects#show` itself: it is a `:public => true` permission, which is not an
-  exemption, because `Role#allowed_permissions` intersects the public
-  permissions with the token's scopes too (`scope.present? ? unscoped & scope
-  : unscoped`), so a token scoped only `view_issues` is refused by Redmine
-  after this server had already admitted the call. The scope map justified the
-  empty entry with a comment about `/custom_fields.json` being admin-gated, an
-  endpoint this tool has never called.
+- `list_project_issue_custom_fields` requires the `view_project` scope instead
+  of none, for the `projects#show` request it makes. `view_project` being a
+  `:public => true` permission is not an exemption, because
+  `Role#allowed_permissions` intersects the public permissions with the
+  token's scopes too (`scope.present? ? unscoped & scope : unscoped`), so a
+  token carrying no scopes is refused by Redmine after this server had already
+  admitted the call. Nothing beyond `view_project` is required: every released
+  Redmine renders the `issue_custom_fields` include gated on
+  `include_in_api_response?` alone, with no permission check. The scope map
+  previously justified an empty entry with a comment about
+  `/custom_fields.json` being admin-gated, an endpoint this tool has never
+  called.
 - `list_redmine_projects` now returns the six fields Redmine's project index
   renders and this serializer discarded: `homepage`, `parent`, `status`,
   `is_public`, `inherit_members` and `updated_on`. `status` is Redmine's
--- a/README.md
+++ b/README.md
@@ -178,6 +178,8 @@ When `REDMINE_AUTOFILL_REQUIRED_CUSTOM_FIELDS=true`, `create_redmine_issue` retr
 - the Redmine custom field `default_value`, or
 - `REDMINE_REQUIRED_CUSTOM_FIELD_DEFAULTS`
 
+In practice only the second one can fire. The server reads project custom fields from `GET /projects/{id}.json?include=issue_custom_fields`, which Redmine renders as id and name only, so it never sees `default_value` -- see [`list_project_issue_custom_fields`](docs/tool-reference.md#list_project_issue_custom_fields). Set the env map if you want autofill to have anything to work with.
+
 Example:
 
 ```bash
--- a/docs/tool-reference.md
+++ b/docs/tool-reference.md
@@ -348,7 +348,7 @@ many requests as it needs.
 
 List the ids and names of the issue custom fields enabled for a project.
 
-**OAuth scopes:** `view_project` and `view_issues`. Both are required: `view_project` for `projects#show`, and `view_issues` for the include, which Redmine renders only for a caller holding it. `view_project` being a `:public => true` permission is not an exemption -- `Role#allowed_permissions` intersects the public permissions with the token's scopes too.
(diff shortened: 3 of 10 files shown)


jztan (changes requested):
@mmahmed The core of this is right and verified: I live-tested on 6.1.1 and 7.0.0 sandboxes and the include really carries only `id` and `name`, the required field that used to report `is_required: false` now reports `null`, and the `TRACKER_BINDINGS_UNREADABLE` refusal works. One blocking issue: the `view_issues` half of the scope-map change. I minted a Doorkeeper token scoped `view_project` only on the 6.1.1 sandbox and the full `issue_custom_fields` include came back, and `projects_helper.rb` in both versions gates the block only on `include_in_api_response?`, not on any permission. So requiring `view_issues` would deny the call, and hide the tool from `tools/list`, for tokens Redmine itself serves, and the gating claim is also stated as fact in the changelog and tool-reference. Could you reduce the entry to `frozenset({"view_project"})` and reword those three places, or share evidence for the gating if I missed it? Two small nits for the same pass: the docstring cross-reference names `_tracker_bindings_readable` but the function is `_tracker_bindings_unreadable_reason`, and one modified tool-reference line carries an em dash where house style is `--`. This will also need a rebase once #239 lands since both touch tools/projects.py.

mmahmed:
You're right, and thank you for live-testing it rather than taking the claim at face value. I've reduced the entry to `frozenset({"view_project"})` and reworded all three places. Pushed as `81f8766`, rebased onto current `develop`.

I want to say exactly how I got it wrong, because the reason turns out to matter for this code.

I checked against a local `6.1-stable` **branch** checkout, not a release. That working copy reports `6.1.3` in `lib/redmine/version.rb`, so it looked like the version I was claiming. But `6.1-stable` and `master` now carry:

```ruby
end if include_in_api_response?('issue_custom_fields') && User.current.allowed_to?(:view_issues, project)
```

while every released version has the bare form you observed:

```ruby
end if include_in_api_response?('issue_custom_fields')
```

I re-checked this against the actual tags rather than my checkout — 6.1.1, 6.1.2, 6.1.3 and 7.0.0 all render the include with no permission check, and only the two branches gate it. So your sandbox result was correct for every Redmine anyone is running, and my "Redmine gates this" was describing an unreleased branch without realising it. (I'd also briefly convinced myself the gate was three days old, from `git log` on that checkout — it's a shallow clone with a single commit, so every line blames to it. Twice wrong from the same working copy.)

Worth flagging as a fact rather than an ask: since the gate is on `6.1-stable` and `master`, a future release will start omitting the array for a caller without `view_issues`, and this tool will then return `[]` for it — the same "absent read as empty" shape this PR exists to fix, arriving from a new direction. `view_project` is still the right entry when that lands, because the tool should stay callable and report the absence rather than be refused; but the array-versus-`null` handling will need to grow a case. Not in this PR.

Changes in `81f8766`:

- `oauth_scopes.py` — entry is `frozenset({"view_project"})`, with the comment now recording that every released Redmine gates the include on `include_in_api_response?` alone, and that requiring more would refuse calls Redmine serves and hide the tool from `tools/list`. The `view_project` half is unchanged and still needed for `projects#show`; `:public => true` isn't an exemption, since `Role#allowed_permissions` intersects public permissions with the token's scopes too.
- `CHANGELOG.md` and `docs/tool-reference.md` — the `view_issues` gating claim removed from both and replaced with the released-version behaviour.
- One more correction you didn't ask for, from the same verification: all three places said `projects_helper.rb` is "unchanged in 6.1, 7.0 and trunk". What's actually unchanged is the two rendered keys — the file itself has moved on trunk, exactly as above. They now say the narrower, true thing.
- Both nits: the cross-reference is `_tracker_bindings_unreadable_reason`, and the `docs/tool-reference.md` line uses `--`.

Rebased onto `develop` at `03f8ff1`, after #235, #237 and #239 landed. Three conflicts, all resolved rather than taken from one side: `CHANGELOG.md` twice (kept both sets of entries, and let this branch's later revision of its own bullet supersede the earlier one), and the `_serialization` import list in `tools/projects.py`, which needed both `_pagination_info` from #239 and `_payload_attr` from here. The `list_redmine_projects` docstring paragraph about custom field definitions also overlapped with #239's trimmed wording; I kept #239's length and folded this branch's point into it, rather than re-expanding a docstring that ships in every `tools/list`.

`2051 passed`, flake8 and black clean locally.

jztan (approved):
@mmahmed Re-verified on `81f8766`: 2051 tests pass locally, and the live checks still hold on both sandbox versions, the six nulls and the tracker refusal. The scope entry now matches what released Redmine serves. Thanks for tracking the claim down to the unreleased branch, and for the heads-up that the gate is coming; I agree `view_project` stays correct when it lands, and the array-versus-null handling can grow a case at that point. Approving and merging.

jztan:
Thanks @mmahmed! Merged. Chasing the include gate down to the unreleased branch was above and beyond; that heads-up will save someone a debugging session when it ships in a release.

PR #235, opened 2026-08-24
feat(contacts): report pagination metadata on a contact list

### LINKED ISSUE #234: `manage_contact` `list` returns no pagination metadata, discarding the `total_count` Redmine sent beside the contacts
## Problem Statement

`manage_contact` with `action="list"` returns a bare array and no pagination metadata at all — no total, no count, no pagination block. A caller cannot tell a complete read from a truncated one, and the only way to find the end of a collection is to keep paging until an empty page comes back, which costs an extra request to learn something the response could have stated.

`list_redmine_issues` already solves this, in the same codebase, with `include_pagination_info=True` returning `{"issues": [...], "pagination": {...}}`. Contacts have no equivalent.

The number is already in the response the connector parses. `_list_contacts_action` does not go through python-redmine's resource manager — it calls the engine directly and gets the decoded JSON envelope:

```python
payload = client.engine.request("get", url, params=params)
raw = payload.get("contacts", []) if isinstance(payload, dict) else []
return [_contact_to_dict(c) for c in raw[:limit]]
```

`payload` is the whole envelope. `payload.get("contacts", [])` keeps the array and discards `total_count`, `offset` and `limit`, which sit beside it. So this is not a missing API capability — it is metadata being parsed and then dropped.

## Proposed Solution

Add `include_pagination_info: bool = False` to the `list` action, mirroring `list_redmine_issues` exactly: bare list when false, so nothing existing changes, and `{"contacts": [...], "pagination": {...}}` when true, with the same eight keys that tool returns — `total`, `limit`, `offset`, `count`, `has_next`, `has_previous`, `next_offset`, `previous_offset`.

The CRM plugin does render the metadata. `app/views/contacts/index.api.rsb` opens with:

```ruby
api.array :contacts, api_meta(:total_count => @contacts_count, :offset => @offset, :limit => @limit) do
```

confirmed against 4.4.5 Pro.

Three details I would suggest doing differently from the issues path, the first two because this endpoint hands over a real total and that one does not, the third because the total it hands over does not always describe the page:

**`has_next` can be measured rather than guessed.** The issues path infers it from `len(result_issues) == limit`, which claims a further page whenever the collection size is an exact multiple of the page size. Given a total that describes the collection being paged, a further page exists exactly when `offset + limit` is below it. Worth keeping the guess as a fallback for a build that renders no `total_count` — but not as `null`, because `null` is falsy and a caller writing `if pagination["has_next"]` would read it as a last page and stop mid-collection. An unknown total should still leave `total` null; it should not make `has_next` lie.

**Report the window Redmine says it applied, not the one requested.** `total_count`, `offset` and `limit` arrive together, so if a build serves a smaller page than asked for, deriving `next_offset` from the requested limit steps straight over the rows it withheld.

**A `search` has to forfeit the total, because the plugin does not narrow it.** This one is easy to miss and it inverts the benefit above, so it is worth stating plainly. `ContactsController#index` counts with `@query.object_count`, which is `objects_scope` with no arguments, while the rows come from `results_scope(:search => params[:search], ...)` — and `ContactQuery` registers no `search` filter, so the search reaches the scope only through that second call. The reported `total_count` therefore describes the collection *before* the search:

```ruby
@contacts_count = @query.object_count                    # no :search
@contacts = @query.results_scope(:search => params[:search], :limit => @limit, :offset => @offset)
```

So on a search matching 3 contacts out of 250, `offset + limit < total` reads `0 + 100 < 250` and claims a further page; at offset 100 it claims another. Three requests to read three rows, and `has_next` true on two responses that returned nothing — where the full-page inference would have said "done" correctly on the first. A searched page is best reported with `total` null, which falls through to that inference. Every other narrowing parameter the tool sends (`project_id`, `tags`, `assigned_to_id`, `author_id`, and the Pro attribute filters) *is* a registered `ContactQuery` filter, so those are counted and the total stays exact — `search` is the only one that is not.

## Use Case

Any read of a contact collection large enough to page. A client that asks for contacts and gets 100 rows currently has no way to know whether that is the whole set or the first page of 250 — so it either always pages to exhaustion, spending requests to discover an end that was already known server-side, or it silently works from a partial set.

It matters most where the caller is an LLM: a truncated list that looks complete produces a confidently wrong answer ("there are 100 contacts"), and nothing in the response contradicts it.

## Feature Category

Existing Tool Enhancement

## Priority

Important

## Alternatives Considered

Paging until an empty page. Works, and is what a caller must do today, but it costs one extra round trip per collection to learn a number the first response carried.

Returning the envelope unconditionally instead of behind a flag. Rejected — it would change the return type for every existing caller of the `list` action. The opt-in flag is also what `list_redmine_issues` already does, so following it keeps the two tools readable the same way.

Omitting `total` entirely when a build does not report it, as `search_redmine_issues` does. Rejected in favour of a stable key set, so a caller reads one shape from both tools and can distinguish "not reported" (`null`) from a number.

## Additional Context

Worth knowing for the "what if there is no `total_count`" branch, because it is a real mode rather than a defensive hypothetical: Redmine's `api_meta` returns `nil` — dropping `total_count`, `offset` and `limit` from the response entirely — whenever `nometa` is present in the request or the `X-Redmine-Nometa` header is set. It is a compatibility mode for ActiveResource clients that cannot deserialize an array carrying attributes.

```ruby
def api_meta(options)
  if params[:nometa].present? || request.headers['X-Redmine-Nometa']
    # compatibility mode for activeresource clients that raise
    # an error when deserializing an array with attributes
    nil
  else
    options
  end
end
```

`app/helpers/application_helper.rb` on `6.1-stable`. The connector sends neither, so the metadata normally arrives — but the branch is reachable, which is why `total` needs a defined meaning when it is missing rather than being assumed present.

## Contribution

mmahmed:
## Summary

Fixes #234. `manage_contact` with `action="list"` returned a bare array and no pagination metadata, so a caller could not tell a complete read from a truncated one. `include_pagination_info=True` now returns `{"contacts": [...], "pagination": {...}}`, the same envelope and the same eight keys `list_redmine_issues` already returns. Default is unchanged — a bare list — so nothing existing moves.

The number was already in the response. `_list_contacts_action` calls the engine directly rather than going through a resource manager, so `payload` is the decoded envelope and `payload.get("contacts", [])` was keeping the array and discarding the `total_count`, `offset` and `limit` beside it. This reads what was already parsed; it adds no request.

## Why the shape matches `list_redmine_issues`

Same opt-in parameter name, same container-keyed-by-resource envelope, same eight keys — `total`, `limit`, `offset`, `count`, `has_next`, `has_previous`, `next_offset`, `previous_offset`. A caller that has learned one tool's pagination has learned this one's. The key set is pinned in a test so it cannot drift.

Two places this deliberately behaves better than the issues path, both because this endpoint hands over a real total and that one does not:

**`has_next` is measured, not inferred** — where the total measures the page. The issues path uses `len(result_issues) == limit`, which reports a further page whenever the collection size is an exact multiple of the page size. With a total in hand, a further page exists exactly when `offset + limit` is below it. A test pins the boundary the inference gets wrong (total 200, limit 100, offset 100 → `has_next` false). `issues.py` is untouched, so no existing tool changes shape.

**The reported window is the one Redmine says it applied**, not the one requested. `total_count`, `offset` and `limit` arrive together, so if a build serves a smaller page than asked for, deriving `next_offset` from the requested limit would step over the rows it withheld.

## The one case where the total must not be believed

`search` is the exception, and it is worth its own section because it inverts the benefit above.

`ContactsController#index` counts with `@query.object_count` — `objects_scope` with no arguments — while the rows come from `results_scope(:search => params[:search], ...)`, and `ContactQuery` registers no `search` filter. So the search narrows the returned page and not the reported count:

```ruby
@contacts_count = @query.object_count                    # no :search
@contacts = @query.results_scope(:search => params[:search], :limit => @limit, :offset => @offset)
```

Deriving `has_next` from that total walks the caller through pages that do not exist. On a search matching 3 of 250 contacts, `0 + 100 < 250` claims a further page; at offset 100 it claims another; three requests to read three rows, with `has_next` true on two responses that returned nothing. The plain full-page inference would have been right on the first.

So a searched page reports `total: null` and falls through to that inference, which is correct for a searched page. `count` still says what the call matched. Every other narrowing parameter the tool sends — `project_id`, `tags`, `assigned_to_id`, `author_id`, the Pro attribute filters — *is* a registered `ContactQuery` filter and so is counted, which is why `search` alone forfeits the total. Checked against redmine_contacts 4.4.5 Pro; tests cover the searched short page, the searched empty page, the searched full page, and that each of the other narrowing parameters keeps its total.

## When a build reports no `total_count`

The CRM plugin does render the metadata — `app/views/contacts/index.api.rsb` wraps the array in `api_meta(:total_count => @contacts_count, :offset => @offset, :limit => @limit)`, checked against 4.4.5 Pro — but the absent case is handled rather than assumed away, since what a given build renders is the build's own business.

`total` is then `null`, which says "not reported" without inventing a number. `has_next` is **not** null: it falls back to the issues-path full-page inference. That is deliberate. `null` is falsy, so a caller writing `if pagination["has_next"]` would read an unknown as a last page and stop mid-collection — silently, which is the failure mode this whole change exists to remove. A full page reported as "ask again" costs one wasted request at worst.

The key set never changes, so the envelope's shape does not depend on the payload. `offset`, `has_previous` and `previous_offset` follow `offset` alone and never depend on the total. `count` is the contacts actually returned, and `limit` is the window Redmine reported applying, falling back to the requested value when it reported none — so those two describe what happened rather than what was asked for, which is the point, but neither is a restatement of the request.

## Changes

`_pagination_info(*, limit, offset, count, total=None)` and `_payload_int` go in `_serialization.py`, whose module docstring already advertises pagination helpers and had none — those keys were being hand-built at four sites in `issues.py` — two of them without `total`, the search API not reporting one — and putting a fifth spelling in `contacts.py` seemed the wrong direction. `_payload_int` rejects `bool` explicitly, since `isinstance(True, int)` is true and a boolean page number is not one. `_contact_pagination_info` stays in `contacts.py` and is now just the part that is contacts-specific: reading `total_count`, `limit` and `offset` off this endpoint's envelope and deciding whether the total may be believed. `issues.py` is untouched, so nothing existing changes shape; it can adopt the helper whenever its own `has_next` is revisited.

`include_pagination_info` is threaded through both the action and the `manage_contact` signature, because the dispatcher hands every parameter to every action. It is also named in `_CONTACT_OWNED_QUERY_KEYS`: without that, `filters={"include_pagination_info": true}` was forwarded to Redmine as a query key it ignores, and the caller got the bare list they had asked to be an envelope — a silent no-op in both directions.

The parameter description carries the contract — the key set, that `total` is `null` whenever no reported total measures the collection being paged, that `search` is one such case, and that `has_next` is always a boolean and so safe to loop on. That is in the `Args:` block rather than under `Returns:` on purpose: FastMCP exposes only the text above `Args:` plus the per-parameter descriptions, so anything a caller needs has to live in one of those two places or it never reaches the model. The mechanism behind the `search` case lives in `docs/tool-reference.md` instead, since the parameter description is shipped in every `tools/list` and a caller only needs the rule.

## Tests

New `tests/test_contact_list_pagination.py`, 37 tests. Fixtures build the real decoded envelope dict rather than a `Mock`, and the module docstring says why: a `Mock` answers `total_count` whether the payload carried one or not, which is the one thing these tests are about.

Covers the default staying a bare list, the envelope shape, the key set pinned, `total` sourced from `total_count`, first/middle/last/past-the-end/short-page/exact-multiple boundaries, `count` following the local slice, a missing total, a non-integer and a boolean total, a non-dict payload, and three error paths. Two tests pin the no-total fallback specifically: a short page with no total is still the last one, and `has_next` is never the falsy `null` a caller would stop on. Four cover the searched page, including that each of the other narrowing parameters keeps its total. One pins the request count at 1, since "the total costs no extra request" is the argument for reading it off this envelope at all and nothing else would catch a second call creeping in.

`1884 passed, 0 failed` on the full non-integration suite, against a `1847 passed` baseline I measured on a clean `origin/develop` worktree rather than assuming it. `flake8 src/ --max-line-length=88` clean, and `black --check` clean on every file this branch touches. Three test files unrelated to this change (`test_env_loading.py`, `test_release_script.py`, `test_wrap_policy.py`) are already unformatted on `develop`; I have left them alone rather than sweeping them into this diff.

## Compatibility

Additive. The flag defaults to false and the default return value is byte-identical to today's, so no existing caller changes. `manage_contact` already validated `limit` (`Field(ge=1, le=100)`) and `offset` (`Field(ge=0)`); neither moves.


jztan (approved):
@mmahmed Verified live against a 6.1.1 sandbox with seeded contacts: `search` returned 1 row with `total_count` 3, so the count mismatch you describe for 4.4.5 Pro holds on the Light build too, and offset pages the searched collection, which confirms the model behind nulling the total. Envelope math checked at all boundaries. Three nits: (1) the added docs/tool-reference.md lines use em dashes, house style is `--`; (2) the retained `search` parameter docstring still says paging a search is "unreliable for more than one page", which now understates what this PR correctly delivers, worth reconciling; (3) optional hardening, derive `next_offset` from `min(applied_limit, requested_limit)` so a server echoing a larger limit than requested can never direct a caller past rows the tool sliced off. Approving; happy to handle these on merge if you prefer.

jztan:
Thanks @mmahmed! Merged. The live sandbox run matched every boundary your tests pin, including the searched-page fallback.

PR #237, opened 2026-08-24
feat: return the caller's project memberships from get_current_user

### LINKED ISSUE #236: `get_current_user` cannot return the caller's memberships, so "which projects do I hold" costs one call per project
## Problem Statement

`get_current_user` cannot return the caller's project memberships, so there is no way to ask "which projects do I hold, and with which roles" in one call.

The tool takes no arguments (`tools/enumeration.py`), calls `user.get("current")` with no `include`, and hand-builds eight keys: `id`, `login`, `firstname`, `lastname`, `mail`, `admin`, `created_on`, `last_login_on`. Memberships are not among them and nothing else on the tool surface answers the question for the current user. `list_project_members` answers it one project at a time, via `project_membership.filter(project_id=...)`, so the cost of answering it today is **one call per candidate project** — 28 on the instance where this came up, and it grows with every project added.

Redmine already sends this data on the endpoint the tool already calls. It is one query parameter away.

## Proposed Solution

Pass `include=memberships` on the existing `user.get("current")` call, behind an opt-in flag in the style of the existing `include_custom_fields` and `include_pagination_info` parameters, and surface each entry with Redmine's own keys — `id`, `project` as `{id, name}`, and `roles` as `{id, name}` with Redmine's `inherited` flag preserved where it sends it.

Three things I would flag for whoever picks this up, all checked against Redmine 6.1-stable, 7.0-stable and master (`app/views/users/show.api.rsb` is byte-identical in all three):

**It works for a non-admin.** The memberships block is gated on `include_in_api_response?('memberships') && @memberships` and nothing else. The action is reachable too: `before_action :require_admin, :except => :show`, `accept_api_auth :show`, and `find_user(false)` resolves `params[:id] == 'current'` behind a bare `require_login`. Redmine also filters the list by `Project.visible_condition(User.current)`, which for `current` is a no-op against their own memberships. python-redmine declares `_includes = ['memberships', 'groups']` on `User` and maps `memberships` to `ProjectMembership`, so the library side is already in place.

**`include` should not be a free-text passthrough.** python-redmine accepts `groups` in the same `_includes` list, but Redmine gates that block on `User.current.admin?`. A non-admin asking for groups therefore gets the key silently omitted, which is indistinguishable from "belongs to no groups" — a wrong answer rather than a refusal. Better to expose memberships specifically, or to handle that gate explicitly.

**The include should go on the original call, not through the lazy attribute.** `memberships` is in `_includes`, so `user.memberships` already resolves today — but `Resource.__getattr__` handles a miss by calling `self.refresh(itself=False, include=attr)`, a second HTTP request. Passing the include up front is one round trip instead of two. This is the same pattern as the issue includes that are read lazily elsewhere in the codebase.

## Use Case

Working out what a person's role is on a deployment whose role names do not map onto the work being done. Who somebody is has to be inferred from what they hold — which projects, with which roles — and that is exactly a memberships read.

Today an agent answering "what am I allowed to work on" walks every visible project with `list_project_members` and filters client-side. That is 28 calls on this instance to produce a few rows, and it is the single largest avoidable call cost in the client that hit it. With memberships on the current-user call it is one call.

It also matters for telling *holding* a project from merely *seeing* one, which `list_redmine_projects` cannot currently express either.

## Feature Category

Existing Tool Enhancement

## Priority

Important

## Alternatives Considered

`list_project_members` per project, which is what the client does now: correct, and linear in the number of projects.

Reading `user.memberships` and letting python-redmine lazily fetch it. This works without any signature change, but it silently costs a second HTTP request per call, so it is the wrong shape for a fix.

`GET /memberships.json` as a cross-project read — Redmine has no such endpoint; memberships are only exposed per project or nested under a user.

## Additional Context

One incidental bug found while checking this, worth fixing in the same change: the docstring says "Resolves to `GET /my/account.json` under the hood." It does not — python-redmine's `User.query_one` is `/users/{}.json`, so the call is `GET /users/current.json`. This is load-bearing rather than cosmetic: `app/views/my/account.api.rsb` renders no memberships block at all, so if the docstring were accurate the feature requested here would be impossible on that endpoint.

One limit of the feature itself, worth knowing before it is built rather than after: Redmine narrows `@memberships` with `Project.visible_condition(User.current)`, and `Project.allowed_to_condition` limits visibility to `status IN (ACTIVE, CLOSED)`. So the list covers active and closed projects but never archived ones, and an empty list means "none visible", not necessarily "none held". That belongs in the tool description, since a caller cannot tell the two apart from the response.

## Contribution

mmahmed:
## Summary

Fixes #236. `get_current_user` took no arguments and could not return the caller's memberships, so "which projects do I hold, and with which roles" cost one `list_project_members` call per project. `include_memberships=True` now answers it in the same request that was already being made — the include rides the existing `GET /users/current.json`, so it adds no round trip. Default is unchanged, so no existing caller moves.

Each entry is `{id, project: {id, name}, roles: [{id, name}]}`, using Redmine's own keys. A role held through a group carries `inherited: true`; Redmine omits the key for a directly held role and so does this, rather than asserting a `false` Redmine never sent.

## Why this is safe for a non-admin token

Checked against 6.1-stable, and 7.0-stable and master are byte-identical for both files, so none of this is version-contingent.

`app/views/users/show.api.rsb` gates the memberships block on `include_in_api_response?('memberships') && @memberships` and nothing else. The action is reachable: `UsersController` has `before_action :require_admin, :except => :show` and `accept_api_auth :show`, and `find_user(false)` resolves `params[:id] == 'current'` behind a bare `require_login`. The `visible?` check in `show` passes too — `Principal.visible`'s non-admin branch matches `id = user.id`, commented "self and members of visible projects". Redmine also narrows `@memberships` with `Project.visible_condition(User.current)`, which for `current` is a no-op against their own memberships.

Worth stating explicitly because two comments in this repo asserted the opposite — that `user.get('current')` "requires admin rights on some Redmine setups". That is not true of stock Redmine, and both comments are corrected here to say what actually keeps those two call sites on httpx.

## Why a boolean and not an `include` passthrough

python-redmine's `User._includes` also accepts `groups`, but Redmine gates that block on `User.current.admin?`. A non-admin asking for groups would get a 200 with the key simply missing — indistinguishable from "belongs to no groups", which is a wrong answer rather than a refusal. Exposing `memberships` specifically avoids handing callers a footgun, and it matches the established split in this codebase: per-include booleans on the python-redmine-backed core tools (`include_watchers`, `include_relations`, `include_custom_fields`), free-text `include` only on the plugin endpoints reached through `engine.request`.

It also cannot proliferate. `User._includes` is exactly `['memberships', 'groups']`, so with `groups` deliberately out, this tool is at its terminal state with one flag.

## The include goes on the original call, not through the lazy attribute

`memberships` is in `_includes`, so `user.memberships` already resolves — but `Resource.__getattr__` handles a missing payload key by calling `refresh(itself=False, include=attr)`, a second request, and to `/users/<id>.json` rather than `/users/current.json` since it re-resolves from `internal_id`. Reading the payload through the existing `_included_list` helper keeps it at one request unconditionally, and returns the payload shape rather than a `ResourceSet`.

I measured this rather than assuming it: one HTTP request on all three paths — flag on with the key present, flag on with the key absent, and flag off. The eight profile `getattr` reads cost nothing, including `admin`, which Redmine omits for a non-admin and which lands on `ResourceAttrError`, an `AttributeError` subclass the three-argument `getattr` absorbs without triggering a refresh.

## Notes on the serializer

A local serializer rather than reusing `tools/projects.py::_membership_to_dict`, because the two Redmine renderers genuinely differ: the user-show one emits no `user`/`group` (the user is the caller, implicitly), so that helper would return two always-null keys, and it flattens roles in a way that drops `inherited`. Roles go through the shared `_named_ref` for the `{id, name}` half, so the display-name policy from #109 has one home.

## Docs

The caller-facing contract — the entry shape, the `inherited`-key rule, and that an empty list means "none visible" rather than "none held" (Redmine's visibility condition covers active and closed projects, not archived) — is stated **above** the `Args:` block, not under `Returns:`. FastMCP exposes only the text above `Args:` plus the per-parameter descriptions, so anything a caller needs has to live in one of those two places or it never reaches the model.

What is *not* above `Args:` is deliberate too. The derivation behind "not admin-only" — `UsersController` exempting `show` from `require_admin`, `find_user(false)` resolving `current` behind a bare `require_login` — is maintainer rationale a caller cannot act on, and the leading prose ships inlined in every `tools/list`. The claim stays in the docstring; the mechanism sits in a comment beside the call, where `oauth_scopes.py` already records the same thing. That cut the shipped description from 1,100 characters to 964.

The docstring also stops naming `GET /my/account.json` as the endpoint. python-redmine's `User.query_one` is `/users/{}.json`, so the call is `GET /users/current.json`. That mattered rather than being cosmetic: `app/views/my/account.api.rsb` renders no memberships block at all, so on the endpoint the docstring named, this feature would have been impossible.

## Tests

New `tests/test_current_user_memberships.py`. The fakes deliberately avoid `Mock` for the user object: a `Mock` answers every attribute, so a lazy `.memberships` read would pass silently — which is the specific regression these tests exist to catch. The fake records lazy reads instead.

Covers the include actually reaching the client call, the entry shape with Redmine's keys, `inherited` preserved on a role and absent on a direct one, no `user`/`group` keys emitted, a user with no memberships, one request only, the default response staying exactly the eight profile keys, and the error envelope.

`1864 passed, 89 deselected` on the full non-integration suite, from a `1847 passed` baseline on `develop`. `flake8 src/ --max-line-length=88` and `black --check src/ --line-length=88` both clean.

## Compatibility

Additive. The flag defaults to false and the default response is byte-identical to today's eight keys. The scope map entry stays `frozenset()`: the endpoint is not admin-gated, the memberships block carries no permission check, and Redmine already filters the list to the caller's visible projects. `view_members` would be the wrong scope to require — it gates reading *who else* is a member of a project, so demanding it would refuse a call Redmine itself allows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,28 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- `get_current_user` can return the caller's project memberships via
+  `include_memberships=True`, as `[{id, project: {id, name}, roles: [{id,
+  name}]}]` using Redmine's own keys. A role inherited from a group
+  membership carries `inherited: true`, the key being absent for a direct
+  role exactly as Redmine renders it. Previously nothing answered "which
+  projects do I hold, with which roles": `list_project_members` costs one
+  request per project. The include rides the existing
+  `GET /users/current.json` call, so this adds no request and needs no new
+  scope -- Redmine gates neither the endpoint nor the memberships block on
+  a permission, and already narrows the list to projects visible to the
+  caller. `groups`, the other include python-redmine accepts here, is
+  deliberately not offered: Redmine gates that block on the caller being an
+  admin, so a non-admin would get a 200 with the key missing, which is
+  indistinguishable from belonging to no groups
+  ([#ISSUE](https://github.com/jztan/redmine-mcp-server/issues/ISSUE)).
+
+### Fixed
+- `get_current_user` docs said the tool resolves to `GET /my/account.json`.
+  It resolves to `GET /users/current.json`; python-redmine only rewrites the
+  path for the `me` id, not `current`. The distinction matters because
+  `/my/account.json` renders no memberships at all.
 
 ## [2.12.0] - 2026-08-22
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   deliberately not offered: Redmine gates that block on the caller being an
   admin, so a non-admin would get a 200 with the key missing, which is
   indistinguishable from belonging to no groups
-  ([#ISSUE](https://github.com/jztan/redmine-mcp-server/issues/ISSUE)).
+  ([#236](https://github.com/jztan/redmine-mcp-server/issues/236)).
 
 ### Fixed
 - `get_current_user` docs said the tool resolves to `GET /my/account.json`.
--- a/docs/tool-reference.md
+++ b/docs/tool-reference.md
@@ -1661,11 +1661,16 @@ list_redmine_users(name="alice")
 
 ### `get_current_user`
 
-Retrieve the currently authenticated user's profile. Resolves to `GET /my/account.json`, so works for any authenticated user (not admin-only). Useful when a user asks the LLM to do something "for me" — the LLM can call this to resolve the current user's ID.
+Retrieve the currently authenticated user's profile. Resolves to `GET /users/current.json`, and works for any authenticated user (not admin-only) — Redmine exempts the `show` action from its admin filter and resolves the `current` id behind a plain login check. Useful when a user asks the LLM to do something "for me" — the LLM can call this to resolve the current user's ID.
 
-**Parameters:** None.
+With `include_memberships` it also answers "which projects am I a member of, and with which roles" in the same request. The alternative is `list_project_members` once per project.
 
-**Returns:** Dict with `id, login, firstname, lastname, mail, admin, created_on, last_login_on`.
+**Parameters:**
+- `include_memberships` (boolean, optional): Add `memberships` to the response. Default `false`. Costs no extra request — the include rides the same call.
+
+**Returns:** Dict with `id, login, firstname, lastname, mail, admin, created_on, last_login_on`. With `include_memberships`, also `memberships`: a list of `{id, project: {id, name}, roles: [{id, name}]}` entries, empty when the caller holds none. A role inherited from a group membership carries `inherited: true`; Redmine omits the key for a direct role, and so does this tool. Redmine limits the list to projects visible to the caller.
+
+Groups are deliberately not offered as an include: Redmine gates that block on the caller being an admin, so a non-admin would get a success response with the key simply missing — indistinguishable from "belongs to no groups".
 
 **Example:**
 ```json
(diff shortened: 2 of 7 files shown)


jztan (approved):
@mmahmed Reviewed and verified: I checked the one-request claim and the lazy-attribute refresh behavior against python-redmine's source, and live-tested the include on Redmine 6.1.1 and 7.0.0 sandboxes with a non-admin account, including the silent omission of `groups` for non-admins that justifies the boolean flag. The corrected admin-rights comments match the controller source. Two nits, neither blocking: the three added lines in docs/tool-reference.md use em dashes where house style is `--`, and it may be worth a one-line note that the memberships list is uncapped (Redmine does not paginate the include). Approving, I can fix the dashes on merge.

jztan:
Thanks @mmahmed! Merged. The one-request discipline and the recorded lazy-read fakes made this a pleasure to verify.

PR #239, opened 2026-08-24
feat: let list_redmine_projects narrow, measure, and return what Redmine sends

### LINKED ISSUE #238: `list_redmine_projects` cannot be narrowed or measured, drops six fields Redmine sends, and is documented as complete when it is active-only
## Problem Statement

`list_redmine_projects` cannot be narrowed, cannot be measured, and does not return most of what Redmine sends it — and its documentation describes none of that. The whole signature is one argument (`tools/projects.py`):

```python
def list_redmine_projects(
    include_custom_fields: bool = False,
) -> Union[List[Dict[str, Any]], Dict[str, Any]]:
```

The body calls `project.all()` unconditionally, hand-builds five keys per project, and returns a bare array. Four things follow from that, and they are the same gap seen from four sides.

**It cannot be narrowed.** There is no `filters`, no `limit`, no `offset`, no `project_id`. Every call fetches every visible project and the caller filters client-side. The sharp edge is that this makes a Redmine configuration unreachable: a project custom field with **"Used as a filter"** enabled is filterable as far as Redmine is concerned, but this tool has no parameter to send `cf_<id>` through, so the setting has no effect for any MCP caller. The asymmetry is easiest to see with the sibling tool — on a deployment where both a contact custom field and a project custom field are flagged filterable, the contact one narrows correctly through `manage_contact`'s `filters`, while the project one cannot be reached at all. The two are configured identically in Redmine and only one is usable, for a reason that lives in this connector rather than in Redmine.

**It cannot be measured.** The response's only top-level content is the array. There is no `total`, no `count`, no pagination block, so a caller cannot state a collection size without counting rows, and cannot tell a narrowed read from a complete one. `list_redmine_issues` and `search_redmine_issues` both offer `include_pagination_info`, so the inconsistency is inside this server, not between it and Redmine — Redmine's projects index does render `api_meta(:total_count => @project_count, :offset => @offset, :limit => @limit)`, so the number is already in the payload this connector parses and throws away.

**It does not return what Redmine sent.** `app/views/projects/index.api.rsb` renders `homepage`, `parent`, `status`, `is_public`, `inherit_members`, `created_on` and `updated_on`, plus custom values and includes. The serializer keeps `created_on` and drops the other six. `is_public` is the one that hurts most: without it there is no way to tell holding a project from merely being able to see one, which is the whole of a membership test.

**And the documentation says otherwise.** The docstring says "Lists all accessible projects in Redmine" and `README.md` says "List all accessible projects." Both are wrong, and not by a little. `ProjectQuery#initialize` sets

```ruby
self.filters ||= {'status' => {:operator => "=", :values => ['1']}}
```

and `Query#build_from_params` clears filters only when `params[:fields]` or `params[:f]` is present; otherwise the constructor's default survives and merely gains short filters for the parameters that were sent. Sending no `status` therefore does not mean "no status filter" — it means active-only. Closed projects are silently absent from a list documented as complete, which is exactly the kind of sentence an LLM caller has no way to check. Separately, `Project.allowed_to_condition` limits a read to `status IN (ACTIVE, CLOSED)`, so archived projects are unreachable through this endpoint by any parameter.

The four compound. `status` is one of the six dropped fields, so a caller cannot see the status of what they *did* get, let alone infer what was filtered out; and with no `filters` there is no way to ask for the closed ones once they know. Each fix on its own leaves the other three halves in place.

## Proposed Solution

One change to one function, because these are one problem:

1. Accept `filters`, `limit` and `offset`, forwarded to the call — `filters` exactly as `manage_contact` already does, including the guard that makes that safe.
2. Offer `include_pagination_info`, returning `{"projects": [...], "pagination": {...}}` with the key set `list_redmine_issues` already returns. Not a new shape: the same one, so a caller reads every paginated tool the same way.
3. Carry the six dropped fields.
4. Correct the documentation, in the docstring and in `README.md`, and keep Redmine's own default rather than overriding it silently.

Checked against Redmine 6.1-stable, and worth stating because these are the parts that are easy to get wrong:

**The projects API index does honour filters.** `ProjectsController#index` runs `retrieve_project_query` -> `retrieve_query(ProjectQuery, ...)`, whose `api_request?` branch calls `@query.build_from_params(params, ...)`; `format.api` then renders `project_scope`, which is the query's own `results_scope`. So parameters reach the filter builder on the JSON path.

**The shorthand form is what to send.** `Query#build_from_params` iterates `available_filters.each_key` and applies `add_short_filter(field, params[field])`, so `cf_42=value` works as a plain query parameter, the same way it does for issues and contacts.

**And this is why a reserved-key guard matters here too.** In the same method, `params[:fields]` or `params[:f]` triggers `self.filters = {}` and rebuilds only from those, and `params[:query_id]` makes `retrieve_query` take the saved-query branch and ignore everything built from the request. Forwarding any of the three answers HTTP 200 with the wrong set, silently. The helper already factored into `_validation.py` rejects them, including the bracketed `f[]` spellings Rack normalises.

**What is filterable**, from `ProjectQuery#initialize_available_filters`: `status`, `id`, `name`, `description`, `parent_id`, `is_public`, `created_on`, `updated_on`, and `cf_<id>` for project custom fields carrying `is_filter`. Those are project custom fields, a different set from issue custom fields. `status` offers only `1` and `5` — `Query#project_statuses_values` builds from `[active, closed]`, and `ProjectAdminQuery` is what adds archived and scheduled-for-deletion, a class this endpoint does not use.

**The total is trustworthy here, with one exception worth designing around rather than discovering.** `@project_count = @query.result_count` and the rows come from `project_scope(:offset, :limit)` — the same query — so the reported count measures exactly the collection being paged, filters included. That is not true of every paginated endpoint, so it is worth saying explicitly. The exception is a deployment that suppresses API metadata: `api_meta` returns `nil` under `nometa` or `X-Redmine-Nometa`, the response then carries no `total_count`, and python-redmine's `bulk_request` falls back to the first page's length and stops paging. The list is truncated *and* the total agrees with the truncation, so it reads as complete — 5,000 projects arrive as 100 rows reporting 100. The truncation is already there today; adding an envelope is what turns it from silent into asserted, so it belongs in the tool description rather than being guarded against, since nothing reachable from a client tells it apart from a genuinely small instance.

**`parent` is conditional.** It renders only `if project.parent && project.parent.visible?`, so an absent `parent` means either top-level *or* child of a project the caller cannot see. The field is worth carrying, but its documentation must not promise that absence means top-level.

One compatibility note: `project.all()` with no limit currently fetches *every* page, because python-redmine reads `total_count` and pages to completion. So a `limit` parameter has to default to preserving that, rather than quietly starting to truncate existing callers.

## Use Case

Selecting the projects that carry a particular project custom field value — on the instance this came from, that is how a book of work is partitioned, so "the projects tagged X" is the first question any client asks. Today that means fetching all projects and filtering client-side: wasteful, growing with the project count, and silently spending the caller's context window on rows that are about to be discarded. It is also the one case where a Redmine administrator has already done the work — turned "Used as a filter" on — and the connector still cannot use it.

The other three show up in the same session. An LLM caller asked "how many projects are there" answers from the length of whatever it happened to receive. One asked "which of these are closed" cannot answer at all, because `status` is not in the payload and could not have been asked for. And one asked "list all the projects" answers confidently with the active ones, which is the failure that costs the most, because nothing in the response or the documentation marks it as partial.

## Feature Category

Existing Tool Enhancement

## Priority

Important

## Alternatives Considered

**Four separate issues, one per gap.** How this started, and it does not survive contact with the code: they are one function and one docstring, so whichever landed first would force a rebase on the other three, and each would arrive with a story that needs the others to make sense. The documentation fix is only verifiable by a caller once `status` is in the payload, and only actionable once `filters` can ask for it. Splitting them ships halves.

**Fetching everything and filtering client-side**, which is what callers do now. It works, and it is what makes the cost case.

**Naming individual filter parameters** (`status`, `is_public`, `parent_id`, ...) instead of a `filters` dict. Reasonable, and could be added later on top, but it cannot express `cf_<id>` without a parameter per custom field, and `filters` is the shape this codebase already chose for exactly that reason.

**Passing `query_id`** to reuse a saved project query. Attractive but a trap: Redmine's `retrieve_query` prefers a saved query over anything built from the request, so it silently discards every other filter in the same call. It is one of the keys the existing guard rejects.

**Changing the `status` default to return active and closed**, so that "all accessible projects" becomes true rather than the sentence being corrected. Considered and rejected: matching Redmine's own default is the behaviour every other client of this endpoint sees, and changing it would alter what existing callers receive in order to fix a documentation error. Documenting the default and making `status` reachable gets a caller everything without moving anything underneath them.

## Additional Context

`is_public` and `status` are both filterable per the list above and both dropped by the serializer, which is a neat illustration that the read and the narrowing are the same gap from two ends — you currently cannot filter on `status` and also cannot see it.

## Contribution

mmahmed:
## Summary

Fixes #238. `list_redmine_projects` had one argument and returned a bare list of five keys, so it could not be narrowed, could not be measured, and did not carry most of what Redmine's project index renders — while its docstring and the README both called it "all accessible projects", which it has never been. This is one change to one function because those are one problem: `status` is both a filter the tool could not send and a field the serializer dropped, so a caller could neither ask what they wanted nor see what they got.

Four things, in the order the commits land them:

- `filters`, `limit` and `offset`, forwarded to `project.all(**params)`, with `filters` following `manage_contact`'s shape and reusing the shared `_reject_reserved_query_keys` helper rather than reimplementing it.
- `include_pagination_info`, returning `{"projects": [...], "pagination": {...}}` with the key set `list_redmine_issues` already returns.
- The six fields Redmine sends and this serializer discarded: `homepage`, `parent`, `status`, `is_public`, `inherit_members`, `updated_on`.
- The documentation corrected — active-only is stated, with the way to reach closed projects, in the docstring and in `README.md`.

No behaviour changes for an existing caller: every new parameter is optional and the default call returns what it always did.

## The `status` default, since it is the reason the docs were wrong

Redmine's `ProjectQuery#initialize` sets `self.filters ||= {'status' => {:operator => "=", :values => ['1']}}`, and `Query#build_from_params` clears filters only when `params[:fields]` or `params[:f]` is present. So sending no `status` does not mean "no status filter" — the constructor's default survives, and this tool has always returned **active projects only**. That is pre-existing, not introduced here.

The default is kept rather than overridden. Matching Redmine's own default is what every other client of this endpoint sees, and changing it would alter what existing callers receive in order to fix a documentation error; making `status` reachable gets a caller everything without moving anything underneath them. So the tool description now says active-only and documents `filters={"status": "1|5"}`. Only `1` and `5` are offered: `Query#project_statuses_values` builds from `[active, closed]`, and archived and scheduled-for-deletion are added by `ProjectAdminQuery`, which this endpoint does not use. `status` is a `:list` filter whose operators are only `["=", "!"]`, so `"*"` is not a way to ask for everything — it would be read as a literal and match nothing.

## Why `filters` is constrained rather than forwarded freely

This is the part I would most like reviewed. `project.all(**params)` routes parameters through `Project.bulk_decode` → `BaseResource.decode`, which is not purely a query-string builder: its `uploads` branch walks a list of dicts and calls `manager.redmine.upload(attachment.pop('path', ''), ...)` for each, reading a local file and POSTing it before the request being asked for is issued. And Redmine's `ApplicationController#api_key_from_request` prefers `params[:key]` over the `X-Redmine-API-Key` header python-redmine sets, so a forwarded `key` substitutes the configured credential — which matters most in `oauth` mode, where identity is meant to be bound to the request's bearer token.

Neither is reachable through `manage_contact`'s `filters`, which goes to `engine.request` and never touches `decode`; the `key` one does apply there, and is worth a separate look.

So `filters` is constrained two ways, both in `_validation.py` beside the existing guard, and both taking the resource's filter-name set as an argument so `manage_contact` can reuse them:

- **Keys must be registered.** The `ProjectQuery` filter names — `status`, `id`, `name`, `description`, `parent_id`, `is_public`, `created_on`, `updated_on` — plus `cf_<digits>`, and the dotted `cf_<id>.cf_<id>` / `.due_date` / `.status` forms that `add_chained_custom_field_filters` registers. An allowlist rather than a denylist because the two directions are not symmetric: Redmine ignores a filter parameter it does not register, so refusing an unknown key costs a caller nothing they could have used, whereas a parameter that is not a filter at all can still be read by another part of the same request.
- **Values must be a single scalar.** A filter value is one string: `add_short_filter` strips an operator prefix and splits the rest on `|`, so nothing legitimate is a list or a dict. `None` is refused because `requests` drops it from the query string entirely, leaving the collection unnarrowed with a 200. A `bool` is refused too, which is worth calling out: it urlencodes as `True`, and a yes/no filter's values are `"1"` and `"0"`, so it would pass a type check and then match nothing — write `{"is_public": "1"}`.

The reserved-key guard stays as it was; this is additional. `fields`, `f` and `query_id` are still refused because Redmine reads them as the query's own definition and discards everything else in the request. `limit`, `offset`, `include_custom_fields` and `include_pagination_info` are refused too, as keys this signature already names — otherwise `filters` is a second, unvalidated route to a parameter that has just been validated.

## `limit` defaults to fetching everything, and is capped at 1000

`project.all()` with no limit already pages to completion — python-redmine's `bulk_request` reads `total_count`, sets `limit = limit or total_count`, and fetches the remainder in chunks of 100. So `limit` defaults to absent rather than to a number, and no `limit` key is put in the params unless the caller passes one.

`_REDMINE_API_PAGE_CAP` is deliberately *not* applied: the sibling tools that clamp to 100 each issue a single `engine.request`, where 100 is a truthful ceiling, whereas this call pages past it, so clamping would make `limit=250` silently return 100. But a *supplied* limit is worse than no limit, and the schema needs to say so, because `bulk_request` does `limit = limit or total_count` — a supplied limit makes it ignore `total_count` entirely and then build `ceil(limit/100) - 1` further requests unconditionally, with no empty-page check. Measured on a seven-project instance: `limit=1000` fires ten requests and `limit=10000` fires a hundred, every one of them returning the same seven rows. So `limit` is bounded `ge=1, le=1000`, matching the ceiling `list_redmine_issues` already uses, and the parameter description says that omitting it is both the way to get everything and, at `offset=0`, the cheap way — ask for a number you want, not a large one meaning "all". The cheapness is specific to `offset=0`, which the `offset` description now says: with no limit, `bulk_request` resolves the missing limit to `total_count` and pages that many rows *from* `offset` rather than up to the end, so the requests past the end are issued anyway and come back empty. Measured on 250 projects, `offset=240` costs three requests to return ten rows and `offset=300` costs three to return none — the same as reading all 250. The rows are right in every case; this is a cost, and it is python-redmine's rather than this tool's, but the parameter promises something about it so the promise is now accurate and pinned.

## The pagination envelope

Same eight keys as `list_redmine_issues`, from the `_pagination_info` helper in `_serialization.py`, whose module docstring already advertised pagination helpers and had none.

The parameter descriptions are also smaller than they were, which matters because every one of them ships on each `tools/list`. With all five parameters this tool became the largest schema in the server — 4,038 characters against `manage_contact`'s 3,389, the `filters` entry alone at 1,523, which is twice `manage_contact`'s and thirteen times `list_redmine_issues`'. Most of that excess explained *refusals*, which the error messages already explain at the point they happen and which `docs/tool-reference.md` carries in full; the accepted keys, the scalar rule and the operator syntax stayed. The narrowing paragraph also warned that Redmine ignores unregistered filter keys and answers 200 with the collection unnarrowed — true of Redmine, but no longer reachable through this tool, since the allowlist refuses those keys before the request is sent. It is now 3,794: `filters` down 352 and the prose down 126, less the 234 the `offset` correction above had to add. Measured as description plus every parameter description — what a client actually receives in `tools/list`. It is still the largest in the server, by 405 characters; the claim here is that it is no longer disproportionate, not that it is small.

One deliberate difference inside those keys, worth stating rather than leaving to be found: given a total, `has_next` is `offset + limit < total` rather than the issue tools' `len(result) == limit`. The inference reports one more page than exists whenever the collection size is an exact multiple of the page size, and here there is an authoritative total to use instead. `list_redmine_issues` is left alone — changing its `has_next` is its own change, with its own callers to consider.

`total` is Redmine's own `total_count`, read off the resource set the rows already came from, so it costs no extra request — unlike the issue tools, which spend a second `limit=1` query on it. Two details are load-bearing and commented at the call site:

**It is read after the loop, never before.** `ResourceSet.total_count` raises `ResultSetTotalCountError` while the set is unevaluated, and iterating it is what evaluates it. Reading it first would both raise and re-request.

**It is passed, never suppressed.** `@project_count = @query.result_count` with the rows from `project_scope(:offset, :limit)` — the same query — so the count measures exactly the collection being paged, filters included. A test class exists specifically to pin that, because the equivalent is *not* true of every paginated endpoint and the rule must not be carried across by analogy.

**One deployment shape defeats the total, and it is documented rather than guarded.** `api_meta` returns `nil` when `nometa` is in the request or `X-Redmine-Nometa` is set, so the response carries no `total_count`, `limit` or `offset` at all. `bulk_request` then takes its fallback branch, sets `total_count = len(first page)` and stops paging — so the list is truncated at one chunk *and* the total agrees with the truncation, which makes the envelope read as a complete collection. Measured: 5,000 projects come back as 100 rows reporting `total` 100 and `has_next` false. The truncation predates this change; the envelope is what turns it from silent into asserted, which is why it is called out. Nothing reachable from the client distinguishes it from a genuinely 100-project instance — same row count, same total, same single request — so a guard would have to guess, and a guess here is worse than the caveat. It is stated on the parameter, in `docs/tool-reference.md` and in the changelog, and there is a test class pinning the behaviour so that the day a signal for it exists, that is the assertion which changes. `nometa` cannot arrive through `filters`: it is not a registered filter name, so the allowlist refuses it.

With no `limit` there is no page size, so the envelope describes one complete read: the delivered row count stands in as the limit, `has_next` comes out false, and `previous_offset` points back at the start. That last one is a correction rather than a shortcut — `_pagination_info` derives `previous_offset` from the limit, and for an offset past the end the row count is 0, which would hand the caller back the offset they are already on and loop a client walking backwards. The only meaningful predecessor of an unbounded read is the beginning. Named regression test.

## The six fields

From `app/views/projects/index.api.rsb`, whose rendered set is byte-identical on `6.1-stable`, `7.0-stable` and `master`. `parent` is `{id, name}` via the existing `_named_ref`, like every other reference this server returns.

A key the payload did not carry is `null`, never a substituted default. That matters most for `is_public` and `inherit_members`, where a fabricated `False` reads as a setting Redmine never sent; two tests turn red if one is substituted.

`parent` is rendered only `if project.parent && project.parent.visible?`, so `null` means top-level **or** a parent this caller cannot see. The documentation says exactly that rather than promising the first, and a test pins that the two cases are indistinguishable.

`status` is Redmine's integer code — `1` active, `5` closed — which is what makes a `filters={"status": "1|5"}` read legible. Until now a closed project could be fetched but not identified.

## Docs

The `status` default and the silent-ignore warning are **above** the `Args:` block; the allowlist, the value rule and the `limit` cost note are in their parameter descriptions. FastMCP exposes only the text above `Args:` plus per-parameter descriptions, so anything a caller needs has to be in one of those two places to reach the model at all. `docs/tool-reference.md` carries the same plus the accepted-key table, and the README's "List all accessible projects" is corrected.

## Tests

Three new modules — `tests/test_project_list_filters.py`, `tests/test_project_list_pagination.py`, `tests/test_project_list_fields.py` — plus `tests/test_limit_offset_schema.py` for the bounds.

No bare `Mock` where presence or absence is the point: a fake project wraps a real `GET /projects.json` payload and raises for absent keys, and a recording manager captures the params.

The security tests use a spy engine subclassing the real `SyncEngine`, so `ResourceManager.all`, `Project.bulk_decode` and `Redmine.upload` all stay real and `request` is the single funnel. `test_uploads_filter_neither_reads_nor_sends_the_file` asserts the file was never opened and no request was made, *before* asserting the error shape — so a regression fails on the file and the socket, not on a return value. A companion test hands the same payload straight to `client.project.all(uploads=[...])` and shows the file *is* opened and its bytes POSTed, so the absence assertion is not vacuous. Deleting the two guard calls fails 27 tests.

Also covered: every allowlisted key forwarding; the five reserved spellings including the bracketed `f[]`/`fields[]`; owned-key rejection, `include_pagination_info` included, so the flag cannot be smuggled through `filters`; non-dict `filters`; `cf_42` accepted and `cf_abc` refused; operator-carrying values; `limit`/`offset` validation and their schema bounds; the envelope's key set against the issue tools'; first, middle, short-last and exact-multiple-last pages; the unlimited read and the offset-past-the-end case; the request count an unbounded offset actually costs, in three cases, since nothing pinned it before; the total surviving a filtered and a custom-field-filtered read; and `include_custom_fields` unchanged.

Two pinning tests changed rather than moved: `test_default_keys_are_unchanged` in `test_project_list_filters.py` and one in `test_project_custom_field_values.py` both asserted the five-key shape, so both now assert the full rendered set — renamed `test_default_keys_are_the_full_rendered_set`. Worth flagging, since they are the tests that would catch an accidental shape change.

`1987 passed, 89 deselected` on the full non-integration suite, from a `1847 passed` baseline on `develop`. `flake8 src/ --max-line-length=88` and `black --check src/ --line-length=88` both clean.

## Compatibility

Additive. Every new parameter is optional; the default call returns the same bare array over the same pages, and the only change a caller sees without opting in is the six new keys on each entry.

The scope map entry stays `frozenset({"view_project"})` — filtering happens through an unsaved `ProjectQuery` built from request params on the same endpoint, and the only branch of `retrieve_query` that performs a permission check is the `params[:query_id]` one, which is refused.

One merge note, since it is easier to say now than to discover during a merge. This branch adds `_pagination_info` and `_payload_int` to `_serialization.py`, and so does the contacts pagination branch in #234/#235 — the same helpers, the same text, in the same place. `_serialization.py` therefore needs no merge at all in either order: the blob is identical on both branches and in the merged tree, and exactly one copy of each helper survives. `docs/tool-reference.md` auto-merges. `CHANGELOG.md` does conflict, because both branches add a bullet under `## [Unreleased]` at the same offset; the resolution is to keep both. The same is true of #236/#237, which adds a bullet in the same place but shares no source file with this branch at all. Checked with `git merge-tree --write-tree` against both rather than assumed: `CHANGELOG.md` is the only conflicting path in either direction. Happy to rebase this one onto whichever lands first.


jztan (approved):
@mmahmed This is excellent work. I verified the `uploads` and `key` findings against python-redmine's source and Redmine's `api_key_from_request`, and ran the tool end-to-end on 6.1.1 and 7.0.0 sandboxes: the six fields, the envelope, `status` filters, and the allowlist refusals (including `key` and `uploads`) all behave as described. The allowlist matches `ProjectQuery`'s registered filters on both versions. Nits, none blocking: em dashes in the added docs/tool-reference.md prose; `test_an_invisible_parent_is_indistinguishable_from_top_level` builds the same payload twice so it cannot fail independently; the offset-cost and nometa paragraphs in the docstring could be trimmed since tool-reference.md carries them. I'll open a follow-up issue for migrating `list_redmine_issues` to `_pagination_info`, which would fix its exact-multiple `has_next` inference and drop its extra count request. Approving.

jztan:
Thanks @mmahmed! Merged. The `uploads` and `key` analysis was genuinely novel security work; a naive filters passthrough would have shipped both vectors, and the spy-engine tests anchoring the exploit both ways are a model for this kind of guard.

PR #251, opened 2026-08-27
docs: say how list_redmine_issues filters can actually be written

### LINKED ISSUE #250: `list_redmine_issues` cannot express Redmine's operator filter forms, and documents the `filters` route that can as not applying to them
## Bug Description

`list_redmine_issues` offers "find unassigned issues" in its own description and then refuses the only value that expresses it.

Redmine carries a filter's operator *inside* the value. `Query#add_short_filter` detects an operator prefix for the filter's type and splits the remainder on `|`, defaulting to `=` over the split list, so `"56|57"` is either tracker, `"!*"` is "none" and `"!4"` is "not 4". `tracker_id`, `priority_id` and `fixed_version_id` are typed `Optional[int]`, and `assigned_to_id` is `Optional[Union[int, Literal["me"]]]`, so none of those forms can be written on the named parameter — and `!*`, unassigned, cannot be written on it at all.

The escape hatch exists and the documentation closes it. `filters` is merged *after* the named parameters, so a key there overrides the parameter of the same name and every one of those forms works. But the `filters` entry reads "Use this for any filter not listed above", which tells a caller the route does not apply to exactly these keys.

Two adjacent gaps in the same docstring, found while establishing the above:

`filters` says nothing about a filter Redmine cannot read. Redmine applies a filter only if it is registered, and otherwise does nothing, silently — so the response is 200 with the collection unnarrowed, which reads as "everything matched" rather than as an error. A custom field enters the registered set only when it carries "Used as a filter". `list_redmine_projects` documents both facts; `list_redmine_issues`, the busier tool, documents neither.

`limit` says "default: 25, max: 1000" and nothing about cost. Since #241 the limit reaches python-redmine whole and is paged there in chunks of 100, and the follow-up requests are built from the number asked for without checking whether the collection is exhausted — so `ceil(limit / 100) - 1` of them are issued regardless. `list_redmine_projects` gained the equivalent sentence in review of #239; the two tools should not disagree about a cost they now share.

## Steps to Reproduce

Against any instance, with `status_id="*"` on every call so the open-only default cannot skew the comparison, and totals read from `include_pagination_info`:

1. `list_redmine_issues(status_id="*")` — the baseline count.
2. `list_redmine_issues(status_id="*", filters={"assigned_to_id": "!*"})` and again with `"*"`.
3. `list_redmine_issues(status_id="*", tracker_id=<a>)`, again with `<b>`, then `filters={"tracker_id": "<a>|<b>"}`.
4. `list_redmine_issues(assigned_to_id="!*")` — the same value on the named parameter.

## Expected Behavior

Steps 2 and 3 narrow the collection, and the docstring says how to write them.

## Actual Behavior

Steps 2 and 3 do narrow correctly — the two `assigned_to_id` halves sum to the baseline and the two trackers sum to the `|` form, which is the discriminating control, since a *discarded* filter also answers 200 with the unnarrowed collection. So the capability is there.

Step 4 returns `INVALID_ARGUMENTS`, "Input should be a valid integer ... or Input should be 'me'".

So the tool can do it, the parameter cannot, and the parameter's documentation points away from the thing that can.

## Version

2.12.0

## Deployment Method

Docker

## MCP Client

Claude Code (CLI)

## Redmine Version

6.1.1

## Relevant Logs

Not applicable — every call above returns HTTP 200 except step 4, which is refused at the FastMCP boundary before a request is made.

## Additional Context

**The fix should not be to widen the parameter types.** `assigned_to_id` is `Optional[Union[int, Literal["me"]]]` deliberately: #116 tightened it from `Union[int, str]` so the boundary refuses garbage rather than passing it to Redmine and returning an empty list a model reads as "nothing matched", and `test_user_id_typing_schema.py` pins that a bare-string branch is a regression. Accepting free-form strings would sell that back. Documenting the `filters` route costs nothing and keeps it.

Happy to open a pull request — the change is in the docstring's `filters`, `limit` and `assigned_to_id` entries and in `docs/tool-reference.md`, which does not currently list `filters` for this tool at all.

mmahmed:
## Summary

Fixes #250.

`list_redmine_issues` offers "find unassigned issues" in its description and then refuses the only value that expresses it. Redmine carries a filter's operator inside the value and joins alternatives with `|`, which the typed parameters cannot represent — so `!*`, `56|57` and `!4` go through `filters`, which is merged after the named parameters and overrides them. That route already worked and was undiscoverable, because the `filters` entry said it was for "any filter not listed above".

This is a documentation change. The typed parameters are deliberately left alone: #116 tightened them so the boundary refuses garbage rather than handing it to Redmine and getting back an empty list, and widening them again would sell that back for something `filters` already does.

## Changes

- `filters` describes the operator-inside-the-value rule with the four forms a caller actually needs, says that it overrides the named parameter of the same name, and states that an operator a filter's type does not accept is read as a literal rather than erroring.
- `filters` states that a filter Redmine cannot read answers 200 with the collection unnarrowed, and that a `cf_<id>` needs "Used as a filter" on. `list_redmine_projects` already carried both sentences.
- `limit` states the paging cost: above 100 the request is paged in chunks of 100, one request per chunk *asked for*, so a ten-issue project read at `limit=1000` costs ten requests to return ten rows.
- `assigned_to_id` points at `filters` for the operator forms, where it already explains that arbitrary strings are refused.
- `docs/tool-reference.md` documents the `filters` parameter, which it had never listed for this tool, and the `limit` cost. The operator sets each filter's type accepts go here rather than in the docstring.

## Testing

`tests/test_issue_filter_forms.py`, 12 cases. Three assert the shipped schema now carries each fact and fail against the parent; the rest are regression pins on the route itself — each short-filter form reaching Redmine verbatim, `filters` overriding the named parameter, and #116's boundary refusal still firing.

The schema assertions read the registered tool through a FastMCP client rather than the source, since only the leading prose and the `Args:` entries are shipped in `tools/list`, and they collapse whitespace so they do not depend on where the docstring happened to wrap.

Measured cost, from the registered schema rather than read off the source: `list_redmine_issues` goes from 2,538 to 3,717 characters, +1,179 across the whole server. Nearly all of it is the `filters` entry, at 119 → 939. That is the entry a caller cannot act without — the operator rule is the whole fix — and the material that is reference rather than instruction went to `docs/tool-reference.md` instead.

Unit suite green on a clean checkout: 2,212 passed under `uv run pytest -m "not integration" tests/`, the selector CI runs. `black --check src/ --line-length=88` and `flake8 src/ --max-line-length=88` clean.

## Checklist

- [x] `python tests/run_tests.py --unit` passes (activate `.venv` first). `--all` additionally needs a live Redmine, which this workstation does not point at
- [x] No manual version bumps
- [x] Docs updated where behavior changed
- [x] Change works with both local and Docker deployments


jztan (approved):
Thanks @mmahmed. Agreed on leaving the parameter types alone.

I ran the forms against my 6.1.1 sandbox: 13 issues total, `!*` gives 7 and `*` gives 6, `1|2` matches the two trackers summed, and a bogus `cf_99999` comes back 200 with all 13. I also read python-redmine's `bulk_request`, and it does build the extra chunks from the requested limit without checking `total_count`, so the `limit` sentence is right.

jztan:
Merged, thanks @mmahmed. The live check against 6.1.1 matched every form you documented, and the schema tests are a nice way to keep the docstring honest.

PR #256, opened 2026-09-01
feat: expose only allow-listed tools via REDMINE_MCP_ALLOW_TOOLS

### LINKED ISSUE #255: Allow exposing a subset of tools (explicit allow list of tool names)
### Problem Statement

There is no way to expose only a subset of tools. A deployment gets all 45 core tools (2.13.0), even when its clients only ever need a handful.

The existing switches cover different axes:

- `REDMINE_MCP_READ_ONLY` is all-or-nothing on writes
- the plugin flags only affect the 13 plugin-gated tools
- `REDMINE_MCP_EXPOSE_ADMIN_TOOLS` covers a single tool

So an operator cannot say "these tools, and nothing else". Client-side filtering is possible in some MCP clients, but it is per-user and users can change it, so it is no basis for an operator guarantee.

### Proposed Solution

New environment variables:

```
REDMINE_MCP_ALLOW_TOOLS=get_redmine_issue,list_redmine_issues,search_redmine_issues,list_subtasks,get_private_notes,get_current_user
REDMINE_MCP_ALLOW_TOOLS_FILE=<path-to-file-with-one-allowed-tool-per-line>
```

- Comma-separated tool names for `REDMINE_MCP_ALLOW_TOOLS`. Unset means all tools, as today.
- Applies to reads and writes alike, and composes with `REDMINE_MCP_READ_ONLY` rather than replacing it — so a subset that includes a write tool is expressible with read-only off.
- Tools not on the list should disappear from `tools/list`, the way plugin-gated tools do since 2.13.0, rather than only refusing when called.
- An unknown name should produce a startup warning, so a typo does not silently remove a tool the operator believes is enabled.

A list rather than a name pattern, because the tool names are orthogonal to subject area: selecting issue work by matching `issue` misses `list_subtasks`, `get_private_notes`, `get_current_user` and `get_mcp_server_info`, while pulling in `list_redmine_issue_statuses` and `list_project_issue_custom_fields`.

### Use Case

We run one shared instance for a team, behind a TLS reverse proxy with `REDMINE_AUTH_MODE=legacy-per-user`, so every user authenticates with their own Redmine API key. It is currently read-only.

We want to allow exactly issue lookup, search, and nothing else — no time tracking, no wiki, no project administration, no file upload. Not because those tools are dangerous, but because we want the granted capability to be a deliberate, reviewable list that grows as we gain confidence. Later that list will likely include one or two write tools, such as adding a note to an issue, which is why `REDMINE_MCP_READ_ONLY` alone does not cover it.

### Feature Category

Configuration/Setup

### Priority

Critical for my use case

### Alternatives Considered

Currently, there are no reliable alternatives or workarounds this MCP server provides.

### Additional Context

I'm happy to let my LLM create a PR instead of yours if you want 😉

### Contribution

andilem:
Closes #255.

`REDMINE_MCP_ALLOW_TOOLS` names the tools a deployment offers; everything else is hidden from `tools/list` and rejected by `call_tool`. `REDMINE_MCP_ALLOW_TOOLS_FILE` takes one name per line with `#` comments, and is used only when the env var is unset — the same env-var-wins-over-file precedence as `get_secret`. Neither set exposes everything, as before.

Built as you suggested: a second pass in `_tool_allow_list.py`, called from `main.py` right after `apply_plugin_visibility`, using the same FastMCP visibility transforms.

### Your four points

- **Runs after plugin visibility, only ever disables.** It computes the complement and calls `mcp.disable(names=...)`; it never enables. A tool that is both allow-listed and hidden by its plugin flag stays hidden, covered by `test_allow_list_cannot_re_enable_a_plugin_gated_tool`. Because plugin visibility enables by tag, the allow list has to be re-applied after any flag change — `test_reapplying_after_a_flag_flip_keeps_the_list_narrow` pins that.
- **`manage_X(action=...)` granularity.** Documented in the new `docs/tool-reference.md` section and in the README row: whole tools only, per-action control stays `REDMINE_MCP_READ_ONLY`'s job, and the two compose — an allow list containing write tools with read-only off is how selective write access is expressed.
- **Plugin-flag intersection and unknown names in tests.** `tests/test_tool_allow_list.py`, 17 cases: env and file parsing, precedence, unreadable file, enumeration, the intersection in both directions, re-application after a flag flip, unknown-name warning, and an all-unknown list.
- **Docs.** README env table (two rows) plus a pointer under *Available Tools*; new *Tool Allow List* section in `docs/tool-reference.md`, cross-linked with *Read-Only Mode*; CHANGELOG under Unreleased.

### One thing for you to weigh

Enumerating the registered tools needs a synchronous read, because the pass runs at import time next to plugin visibility. FastMCP 3.4.7 has no public synchronous enumeration — `list_tools`, `get_tool` and `_list_tools` are all coroutines — so `registered_tool_names()` reads `mcp._local_provider._components`, whose keys are `"tool:<name>@<version>"`. That is private API. It degrades to a warning and no filtering if the registry ever moves, and `test_registered_tool_names_survives_a_missing_registry` covers that path, but you may prefer one of:

1. an async pass in the lifespan using the public `await mcp.list_tools()` — public API, and the list is already post-plugin-visibility so the complement cannot resurrect anything, at the cost of moving the call site and the test idiom;
2. middleware in the shape of `ScopeEnforcementMiddleware` (`on_list_tools` + `on_call_tool`), which needs no enumeration at all.

Happy to switch to either.

### Two judgment calls

- **A configured but empty list is ignored with a warning**, rather than hiding every tool. Hiding everything cannot be an intended configuration and would look like a broken server; the warning says to remove the variable instead. Say the word if you would rather it be honoured literally, or refuse at startup.
- **All-unknown lists are honoured** and expose nothing, on the grounds that silently ignoring a filter is worse than an empty surface.

### Tests

`2263 passed` on `python -m pytest tests/ -m "not integration"`. Three failures are pre-existing on clean `develop` in my environment and unrelated to this branch — I verified by stashing: two in `test_release_script.py` and one symlink test. All three are Windows-only. The two release-script ones come from `scripts/release.py` and the test itself calling `Path.read_text()` without an encoding, which picks cp1252 here and dies on a UTF-8 byte in README.md; the README Contributors list is in fact already in sync. Untouched in this PR, mentioning it in case it is worth its own issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- `REDMINE_MCP_ALLOW_TOOLS` (and `REDMINE_MCP_ALLOW_TOOLS_FILE`) expose only
+  the named tools; everything else is hidden from `tools/list` and rejected
+  by `call_tool`. The pass runs after plugin visibility and only ever
+  disables, so a listed tool whose plugin flag is off stays hidden. Unknown
+  names are ignored with a startup warning. Granularity is whole tools --
+  per-action control on `manage_X` tools remains `REDMINE_MCP_READ_ONLY`'s
+  job, and the two compose, so an allow list containing write tools with
+  read-only off is how selective write access is expressed.
+  ([#255](https://github.com/jztan/redmine-mcp-server/issues/255))
+
 ### Fixed
 - `get_redmine_attachment` no longer hardcodes `http://` in the download
   `uri` and no longer appends default ports. The scheme now honors a new
@@ -19,6 +30,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Contributors
 - @andilem reported the TLS download URI bug with a precise diagnosis and
   fix proposal ([#252](https://github.com/jztan/redmine-mcp-server/issues/252))
+- @andilem proposed and implemented the tool allow list
+  ([#255](https://github.com/jztan/redmine-mcp-server/issues/255))
 
 ## [2.13.0] - 2026-08-29
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,13 +9,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - `REDMINE_MCP_ALLOW_TOOLS` (and `REDMINE_MCP_ALLOW_TOOLS_FILE`) expose only
-  the named tools; everything else is hidden from `tools/list` and rejected
-  by `call_tool`. The pass runs after plugin visibility and only ever
-  disables, so a listed tool whose plugin flag is off stays hidden. Unknown
-  names are ignored with a startup warning. Granularity is whole tools --
-  per-action control on `manage_X` tools remains `REDMINE_MCP_READ_ONLY`'s
-  job, and the two compose, so an allow list containing write tools with
-  read-only off is how selective write access is expressed.
+  the named tools; everything else is hidden from `tools/list` and refused by
+  `call_tool` with a `TOOL_NOT_ALLOWED` envelope. Enforced by middleware, so
+  the restriction reads no server internals and cannot fail open. It runs
+  alongside plugin visibility and only ever narrows, so a listed tool whose
+  plugin flag is off stays hidden. A variable that is set but names no tool
+  refuses to start; names matching no tool are warned about at startup.
+  Granularity is whole tools -- per-action control on `manage_X` tools
+  remains `REDMINE_MCP_READ_ONLY`'s job, and the two compose, so an allow
+  list containing write tools with read-only off is how selective write
+  access is expressed.
   ([#255](https://github.com/jztan/redmine-mcp-server/issues/255))
 
 ### Fixed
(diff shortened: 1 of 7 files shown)


jztan (changes requested):
Thanks @andilem, the shape is right and the intersection tests pin the property I cared about. One blocker.

The allow list fails open. If FastMCP moves `_local_provider._components`, `registered_tool_names()` comes back empty, `apply_tool_allow_list` returns `None`, and every tool is exposed with nothing but a warning. I tried it with a bare provider object: `REDMINE_MCP_ALLOW_TOOLS=get_redmine_issue` served 41 tools, `delete_redmine_issue` among them. The pin is `fastmcp>=3.0.0,<4`, so a routine lock refresh is enough to trigger it. The same widening happens on `if not registered` and on the empty-list branch, which I think answers your first judgment call: I would rather it refuse to start than expose everything.

I would like you to go with your option 2, the middleware. `on_list_tools` and `on_call_tool` both exist in 3.4.7, so it needs no private API, and it cannot fail open because it filters whatever list it is handed. It also matches `_scope_middleware.py`. Most of your tests assert on `tools/list` output rather than on internals, so they should survive the swap. If you would rather not rewrite that much, I will take raising on all three branches instead of returning `None`, but the middleware is what I want.

Two smaller ones:

- `get_allowed_tools` branches on `raw is None` while `get_secret` branches on truthiness. So an empty `REDMINE_MCP_ALLOW_TOOLS` next to a valid `_FILE` ignores the file and exposes everything, which is what `docker-compose` produces for an unset `${VAR}`.
- `cleanup_attachment_files` is only registered under `REDMINE_MCP_EXPOSE_ADMIN_TOOLS=true`, so allow-listing it with that flag off reports a typo when the name is spelled correctly.

Docs and CHANGELOG look good. On your footnote, your #257 already touches `scripts/release.py` and `tests/test_release_script.py`, so it should cover those two failures. The symlink one is not in that diff though. Could you open a separate issue for it with the failure output and the test name? You hit these Windows cases far more than I do, so I would rather have it tracked in your words than lose it at the bottom of this thread.

jztan (approved):
@andilem this is what I wanted. Verified through the real `main.py` import: one allow-listed tool lists one tool, `delete_redmine_issue` comes back `TOOL_NOT_ALLOWED`, `" , "` refuses to start, unset still serves all 45. 2277 passed locally.

Approving; merging once CI is green on 8883210. The symlink issue is still yours to open when you get a chance.

jztan:
Merged, thanks @andilem. The middleware rewrite and the registry-independence test are exactly the shape I wanted; this ships in the next release.

PR #257, opened 2026-09-01
fix: read and write text files with an explicit utf-8 encoding

andilem:
Text file I/O without an explicit `encoding` falls back to the locale codec. On a Windows machine with a cp1252 locale that breaks the release script:

```
$ python scripts/release.py --sync-contributors
  File "scripts/release.py", line 340, in update_readme_contributors
    content = readme.read_text()
UnicodeDecodeError: 'charmap' codec can't decode byte 0x8f in position 18068
```

`README.md` contains the emoji variation selector `U+FE0F`, which encodes as `EF B8 8F`, and `0x8F` is unmapped in cp1252. `docs/tool-reference.md` has the same byte. Two cases in `test_release_script.py` failed for the same reason, since they also read the committed `README.md` and `CHANGELOG.md` without an encoding:

```
FAILED tests/test_release_script.py::test_committed_readme_contributors_match_the_changelog
FAILED tests/test_release_script.py::test_sync_contributors_flag_is_wired_up
```

Both pass now, and `--sync-contributors --dry-run` runs. Worth noting that the README Contributors list was never actually stale — the failure was purely the decode.

### What changed

- `scripts/release.py`: all 15 `read_text`/`write_text` calls — `pyproject.toml`, `server.json`, `README.md`, `CHANGELOG.md`, the requirements export and the notes files.
- `tests/test_release_script.py`: the fixture writes, and the two reads of the committed `README.md` and `CHANGELOG.md`.
- Attachment metadata I/O in `tools/files.py`, `file_manager.py` and `_http_routes.py`. `json.dump` defaults to `ensure_ascii=True`, so these are latent rather than live — but a metadata file carries user-controlled `original_filename` values, and relying on that default to keep them ASCII is fragile.

Binary-mode opens are untouched; they take no `encoding` argument.

### Not changed

Roughly 40 further `write_text` calls in other test files write ASCII fixtures into `tmp_path`. They cannot hit this, and touching them would bury the fix in noise. Happy to add them if you would rather have the whole codebase consistent — and if you want this enforced, `flake8-encodings` catches the pattern; flake8 alone does not.

### Test run

`2248 passed`, one unrelated pre-existing failure (`test_cert_path_symlink_resolution`, which needs the Windows symlink privilege).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Fixed
+- Text files are now read and written with an explicit `encoding="utf-8"` in
+  `scripts/release.py` and in the attachment metadata I/O. Without it Python
+  picks the locale codec, so on a Windows machine with a cp1252 locale
+  `scripts/release.py --sync-contributors` died on `README.md`: the emoji
+  variation selector `U+FE0F` encodes as `EF B8 8F`, and `0x8F` is unmapped
+  in cp1252. Two `test_release_script.py` cases failed for the same reason.
 - `get_redmine_attachment` no longer hardcodes `http://` in the download
   `uri` and no longer appends default ports. The scheme now honors a new
   `PUBLIC_SCHEME` env var, or is derived (`https` when `PUBLIC_PORT=443`),
--- a/scripts/release.py
+++ b/scripts/release.py
@@ -79,7 +79,7 @@ def run_command(
 def get_current_version(project_root: Path) -> str:
     """Read current version from pyproject.toml."""
     pyproject = project_root / "pyproject.toml"
-    content = pyproject.read_text()
+    content = pyproject.read_text(encoding="utf-8")
     match = re.search(r'^version\s*=\s*"([^"]+)"', content, re.MULTILINE)
     if not match:
         print("Error: Could not find version in pyproject.toml")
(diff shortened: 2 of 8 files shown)


jztan (dismissed):
Thanks @andilem, approving.

I checked the rest of your branch and nothing is left over. Every text file read and write in `src/` and `scripts/` now names an encoding.

Two more need to go in before I merge this. You said the test files you left alone only write plain ASCII into temp folders, which is true. But two of them read rather than write, and not from temp folders. `tests/test_offload.py:194` opens every source file in the package without naming an encoding, and some of those files already contain non-ASCII characters. It gets away with it today because those characters, an em dash and a `✓`, come out as garbled text on a cp1252 machine instead of crashing. The day someone puts an emoji in a docstring, it will crash the same way `README.md` did. `tests/test_oauth_scopes.py:13` reads a checked-in fixture that is all ASCII for now. Please add `encoding="utf-8"` to both, the same fix you already made everywhere else.

You have already named the symlink failure here, so please open the issue I asked for on #256 with that name and a traceback.

jztan (changes requested):
Thanks @andilem, close. One thing before I merge.

I checked the rest of your branch and nothing is left over. Every text file read and write in `src/` and `scripts/` now names an encoding.

Two more need to go in before I merge this. You said the test files you left alone only write plain ASCII into temp folders, which is true. But two of them read rather than write, and not from temp folders. `tests/test_offload.py:194` opens every source file in the package without naming an encoding, and some of those files already contain non-ASCII characters. It gets away with it today because those characters, an em dash and a `✓`, come out as garbled text on a cp1252 machine instead of crashing. The day someone puts an emoji in a docstring, it will crash the same way `README.md` did. `tests/test_oauth_scopes.py:13` reads a checked-in fixture that is all ASCII for now. Please add `encoding="utf-8"` to both, the same fix you already made everywhere else.

You have already named the symlink failure here, so please open the issue I asked for on #256 with that name and a traceback.

jztan:
Merged, thanks @andilem. Both reads are fixed and I could not find any other text I/O without an encoding in src, scripts or tests. The suite passes on a scratch merge with current develop.

PR #260, opened 2026-09-02
fix(tests): skip the cert symlink test where symlinks need a privilege

### LINKED ISSUE #259: test_cert_path_symlink_resolution fails on Windows without the symlink privilege
### Bug Description

`tests/test_ssl_configuration.py::TestCertificateFileValidation::test_cert_path_symlink_resolution` creates a symlink in its setup. On Windows that requires `SeCreateSymbolicLinkPrivilege`, which an ordinary user account does not hold unless Developer Mode is on or the shell is elevated. The call raises `OSError: [WinError 1314]` before the test reaches the path resolution it means to verify, so the failure says nothing about the code under test.

It is the only failure of its kind in the suite and the one red test in an otherwise green local run for a Windows contributor. CI does not catch it, because the Windows job has the privilege available, so it only bites people running the suite locally. A suite that is known-red on a developer machine is how the next real failure gets waved through, which is the actual cost here.

Split out of #256 at your request, so it does not get lost at the bottom of that thread.

### Steps to Reproduce

On Windows, with Developer Mode off and a non-elevated shell:

1. `git clone https://github.com/jztan/redmine-mcp-server && cd redmine-mcp-server`
2. `python -m venv .venv && .venv\Scripts\activate`
3. `pip install -e .[dev]`
4. `python -m pytest "tests/test_ssl_configuration.py::TestCertificateFileValidation::test_cert_path_symlink_resolution"`

Or reproduce the underlying restriction on its own, without the suite:

```python
python -c "import pathlib, tempfile; d = pathlib.Path(tempfile.mkdtemp()); (d / 'target').write_text('x'); (d / 'link').symlink_to(d / 'target')"
```

To confirm the privilege is the cause, the same commands pass in a shell started with *Run as administrator*, or after enabling Settings → System → For developers → Developer Mode.

Environment: Windows 11 Pro 26200, Python 3.14.6, non-elevated shell, `AllowDevelopmentWithoutDevLicense` unset.

### Expected Behavior

Where symlinks cannot be created at all, the behaviour under test is untestable and the case should skip with a clear reason rather than error. `python tests/run_tests.py` should come back green on Windows, and coverage should stay unchanged everywhere symlinks do work, CI included.

The suite already has the pattern one file over, in `test_resolve_local_file_rejects_symlink_escape` (`tests/test_upload_content_resolution.py`):

```python
    try:
        link.symlink_to(target)
    except (OSError, NotImplementedError):
        pytest.skip("symlinks not supported on this platform")
```

`NotImplementedError` is worth keeping in the tuple for the same reason it is there: older Python on Windows raised that instead of `OSError`. These two are the only tests in the suite that create symlinks, so an inline `try` at each site is probably cheaper than a shared helper or a `skipif` marker — worth revisiting if a third ever appears.

Happy to send this as a PR if you would like it.

### Actual Behavior

The test errors in setup:

```
FAILED tests/test_ssl_configuration.py::TestCertificateFileValidation::test_cert_path_symlink_resolution

    def test_cert_path_symlink_resolution(self, tmp_path):
        """Should resolve symlinks to actual certificate file."""
        # Create actual certificate file
        actual_cert = tmp_path / "actual_ca.crt"
        actual_cert.write_text("-----BEGIN CERTIFICATE-----\nFAKE\n")

        # Create symlink to certificate
        symlink_cert = tmp_path / "ca.crt"
>       symlink_cert.symlink_to(actual_cert)

tests\test_ssl_configuration.py:123:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

    def symlink_to(self, target, target_is_directory=False):
        """
        Make this path a symlink pointing to the target path.
        Note the order of arguments (link, target) is the reverse of os.symlink.
        """
>       os.symlink(target, self, target_is_directory)
E       OSError: [WinError 1314] A required privilege is not held by the client:
E       'C:\Users\...\AppData\Local\Temp\pytest-of-alr\pytest-21\test_cert_path_symlink_resolut0\actual_ca.crt'
E       -> 'C:\Users\...\AppData\Local\Temp\pytest-of-alr\pytest-21\test_cert_path_symlink_resolut0\ca.crt'

C:\Users\...\Python\pythoncore-3.14-64\Lib\pathlib\__init__.py:1213: OSError
1 failed in 0.87s
```

The WinError text is localised; the German original reads "Dem Client fehlt ein erforderliches Recht", translated above.

Every other test in `tests/test_ssl_configuration.py` passes.

### Version

develop

### Deployment Method

Source (development)

### MCP Client

Other (specify in additional context)

### Redmine Version

_No response_

### Relevant Logs

```shell

```

### Additional Context

_No response_

andilem:
Closes #259.

`test_cert_path_symlink_resolution` calls `Path.symlink_to` in its setup. On Windows that needs `SeCreateSymbolicLinkPrivilege`, which an ordinary account does not hold outside Developer Mode or an elevated shell, so the test errors with `OSError: [WinError 1314]` before it reaches the resolution behaviour it means to check. Details and the full traceback are in the issue.

The guard matches `test_resolve_local_file_rejects_symlink_escape` in `tests/test_upload_content_resolution.py`, the suite's only other symlink test: create the link inside a `try`, skip on `OSError` or `NotImplementedError`. `NotImplementedError` stays in the tuple for the reason it is there already — older Python on Windows raised that instead of `OSError`. Two call sites did not seem worth a shared helper or a `skipif` marker; worth revisiting if a third appears.

Test-only, plus a CHANGELOG entry. No `src/` change, and coverage is unchanged wherever symlinks can actually be created, CI included. `tests/test_ssl_configuration.py` goes from 19 passed / 1 failed to 19 passed / 1 skipped.

Branched off current `develop` (`0748ebf`), and independent of #257 — I checked both merge directions with `git merge-tree` and neither conflicts, despite both touching the CHANGELOG, so the two can land in either order.

Full suite on Windows: 2286 passed, 3 failed, 99 skipped. The three are unrelated to this change: two are the missing-`encoding` reads in `tests/test_release_script.py` that #257 fixes, and `test_ssl_integration.py::test_ssl_cert_file_exists` wants `tests/fixtures/ssl/ca-cert.pem`, which `generate-test-certs.sh` produces and which is not checked in.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   deployments behind a TLS-terminating reverse proxy get a usable URL
   without config changes.
   ([#252](https://github.com/jztan/redmine-mcp-server/issues/252))
+- `test_cert_path_symlink_resolution` now skips instead of erroring where
+  symlinks cannot be created. Creating one on Windows needs
+  `SeCreateSymbolicLinkPrivilege`, which an ordinary account does not hold
+  outside Developer Mode or an elevated shell, so the test failed in its
+  setup for every Windows contributor running the suite locally. Guarded the
+  same way as `test_resolve_local_file_rejects_symlink_escape`, the suite's
+  other symlink test.
 
 ### Contributors
 - @andilem reported the TLS download URI bug with a precise diagnosis and
--- a/tests/test_ssl_configuration.py
+++ b/tests/test_ssl_configuration.py
@@ -118,9 +118,14 @@ class TestCertificateFileValidation:
         actual_cert = tmp_path / "actual_ca.crt"
         actual_cert.write_text("-----BEGIN CERTIFICATE-----\nFAKE\n")
 
-        # Create symlink to certificate
+        # Create symlink to certificate. Windows needs
+        # SeCreateSymbolicLinkPrivilege for this, which an ordinary account
+        # does not hold outside Developer Mode or an elevated shell.
         symlink_cert = tmp_path / "ca.crt"
-        symlink_cert.symlink_to(actual_cert)
+        try:
+            symlink_cert.symlink_to(actual_cert)
+        except (OSError, NotImplementedError):
+            pytest.skip("symlinks not supported on this platform")
 
         # Resolve the symlink
         cert_path = Path(symlink_cert).resolve()


jztan:
Merged, thanks @andilem. Good bug report on #259, the repro and the elevated-shell check made it a quick review. With #257 in, the suite should be green locally on Windows now.

PR #268, opened 2026-09-06
feat: pass through non-standard top-level issue fields as unmapped_fields

### LINKED ISSUE #263: Issue serializers drop top-level fields added by Redmine distributions and plugins
Some Redmine distributions and plugins add their own top-level keys to the issue JSON.
Easy Redmine, for instance, returns this on every issue:

    "easy_sprint": {"id": 356, "name": "June 2025", "due_date": "2025-06-30"},
    "easy_story_points": 0,
    "is_favorited": false

python-redmine keeps them in the decoded payload, but `_issue_to_dict` and
`_issue_to_dict_selective` build the result from a fixed key set, so they never reach
the client. Today the only way to get them is a second request to `/issues/{id}.json`
outside the MCP server.

I'd like to pass those keys through under a single `extra_fields` sub-object, read from
`raw()` (never `getattr`, which can trigger a lazy fetch on include names), and omit the
key when there is nothing to report so stock Redmine payloads don't change. Same idea as
the `address` pass-through in `_contact_to_dict`: the plugin names those fields itself,
and a hardcoded list drops what it doesn't know.

Would you take a PR for this? I have one ready with tests.

gino8080:
## Summary

Fixes #263

The pass-through you approved in the issue, reworked to the shape you asked for. All six points are in. One of them did not turn out the way you expected, see the note at the end.

## Changes

- The skip set is now the mapped keys plus `set(Issue._includes) | set(Issue._relations)`, read off the resource class instead of a hand-written list. An `include=journals` payload stays out of `unmapped_fields` and cannot sidestep the journal pagination in `get_redmine_issue`.
- `None` values are dropped and the key is omitted when nothing is left, so the seven null include keys on a stock Redmine no longer show up.
- Strings go through `wrap_insecure_content()`, nested ones included. Dict keys are field names, so they stay as they are.
- A value over 1000 characters once serialized is skipped, instead of a per-plugin skip list.
- `_issue_to_dict_selective` only emits the key when the caller names it in `fields`. That was already the behaviour, it just had no test pinning it.
- `total_estimated_hours` and `total_spent_hours` are mapped as first-class keys on both serializers.

One decision I had to make on my own: `tags`. With `REDMINE_TAGS_ENABLED=true` the key has its own serializer, so it is skipped. With the flag off nothing else emits it, so it passes through as an unmapped field. That is also what makes it usable as the assertion target you suggested. Say the word if you would rather have it skipped unconditionally.

## Testing

- `tests/test_issue_unmapped_fields.py`: 27 cases off a fake `raw()` with the Easy Redmine keys. Covers the cap at and over the boundary, wrapping of nested strings and strings inside lists, a populated include staying out, and both branches of the `tags` decision.
- One integration test in `tests/test_integration.py` asserting the shape on a live issue: no keys the serializer already handles, no nulls, nothing over the cap, strings wrapped, and the two rollup fields present.
- Full suite on develop after rebase: 2332 passed, 98 skipped. `black --check src/ tests/` and `flake8 src/ --max-line-length=88` clean.
- Ran live against Easy Redmine 11plus.5.1. `unmapped_fields` came back with `easy_sprint`, `easy_story_points`, `is_favorited` and `css_classes`, `total_estimated_hours` and `total_spent_hours` as mapped keys, and `fields=["id", "unmapped_fields"]` returning exactly those two.

About `css_classes`: on my instance it serializes to 154 characters, so the 1000 character cap does not touch it. The cap is still the better rule than a name list, since a name list only covers the plugins we have already seen. But if the goal was specifically to keep `css_classes` out, size will not do it and you need a different filter.

## Checklist

- [ ] Change works with both local (`uv run python -m redmine_mcp_server.main`) and Docker (`docker-compose up`) deployments. Ran locally only. The change is in the issue serializers, so it does not touch anything deployment specific.


jztan (changes requested):
@gino8080 thanks for this, and for running it against a real Easy Redmine first. That is the
half I can't check here. On my side I put it on both sandboxes against current develop: the
seven null include keys are gone on 6.1.1 and 7.0.0, `total_spent_hours` comes through
mapped, and the unit suite passes. Three things and then I'll merge.

1. The cap is measured before wrapping, so it does not bound what reaches the client. I built
   a value of 38 short string pairs: 978 characters serialized, which passes the cap, then
   6678 once every leaf is wrapped. That is roughly 7x, or 668KB from one plugin field on a
   100 issue listing. Measure the length after wrapping and drop it there.

2. The CHANGELOG entries landed in `## [2.14.0] - 2026-09-05`, which shipped two days ago.
   Move them under `## [Unreleased]`. Leave the Contributors credit to me: adding your handle
   there trips a CI test until I run the sync script.

3. On the sparse fallback in `_hydrate_search_results`, `raw()` still holds `title`, `url`
   and `datetime` from the search endpoint, so a hydration failure reports three stock
   Redmine fields as plugin additions. I reproduced it on the sandbox. Add those three to the
   skip set.

Your `tags` decision is right. The key only shows up on the show endpoint on both my
sandboxes, so nothing is lost with the flag either way. Keep the size rule for `css_classes`
as well, since a name list only covers the plugins I already know about. One small thing: the
changelog says `list_redmine_issues` exposes the key via `fields=["unmapped_fields"]`, but the
default path emits it too.

None of it is much work. Give me a shout when it is pushed and I will get it in.

gino8080:
@jztan pushed as 776d265. All three are in.

1. The cap is now measured on the wrapped value, so it bounds what actually
reaches the client. Your 38-pair case is the new test: it serializes to 978
characters raw, passes, then fails once every leaf is wrapped. Nothing real on my
Easy Redmine instance gets caught by the move. The biggest passing value is
`easy_sprint` at 76 characters raw, 226 wrapped.

2. Changelog entries moved to `[Unreleased]`, and `## [2.14.0]` is byte-identical
to develop again. No Contributors edit from me. I also fixed the line you caught:
it now just says `list_redmine_issues` exposes the key, without implying you need
`fields=["unmapped_fields"]` to get it.

3. `title`, `url` and `datetime` are in the skip set, with a comment pointing at
the sparse-row path in `_hydrate_search_results`. I reproduced it live before
fixing it: on my instance the sparse search rows carry exactly those three keys
beyond the seeded includes, and `unmapped_fields` now comes back empty for them
instead of reporting three stock fields as plugin additions. Good catch, that one
would have been confusing to debug from the client side.

On `css_classes`: keeping the size rule as you said, but note it still passes here
even after the move. It is 79 characters raw and 154 wrapped on my instance, so
the 1000 cap does not touch it. If you want it out, it needs a different rule.

Unit suite: 2330 passed, 31 skipped, flake8 clean. `black --check` flags
`test_wrap_policy.py`, `test_release_script.py` and `test_env_loading.py`, none of
which this branch touches, so I left them alone. Might be a black version drift on
my side.

Live re-verification against Easy Redmine 11plus.5.1 after the changes: issue with
the plugin fields returns `easy_sprint`, `easy_story_points`, `is_favorited` and
`css_classes` under `unmapped_fields`, `total_estimated_hours` and
`total_spent_hours` come through mapped, and `fields=["id", "unmapped_fields"]`
returns exactly those two keys.

jztan:
@gino8080 merged, thanks. Your 38-pair case drops now when I re-run it, and the sparse search
rows come back empty. Both sandboxes are unchanged on stock payloads.

Two small things I'll pick up on develop so you don't need another round. One is the
Contributors credit, which needs the sync script run with it. The other is a clause in
tool-reference.md saying the cap is what keeps `css_classes` out. It doesn't, at 206
characters wrapped against a 1000 cap, exactly as you measured.

The black failures aren't your version, by the way. Those three files fail here on develop
too, so they're mine to clean up.

PR #271, opened 2026-09-07
feat(news): read and manage project announcements

### LINKED ISSUE #269: Add news tools: read and manage project announcements
### Problem Statement

`news` is one of the few core Redmine resources with no coverage at all, and there is no workaround inside the server:

- There is no list or get tool for news.
- `search_entire_redmine` cannot reach them either. It hard-codes `allowed_types = ["issues", "wiki_pages"]` (`tools/search.py:127`) and says so in its docstring, even though the module docstring advertises "issues, projects, wiki pages, news, etc.".

So a client asking "what was announced in this project" comes back empty, and the only place to read it is the web UI. Teams put release notes, maintenance windows and rollout notices there — exactly the context an assistant needs when it summarizes a project's state or answers "is anything going on with X". A project status summary that silently omits "we are migrating the database on Saturday" is worse than no summary.

python-redmine has the resource fully wired already (`News`, read since Redmine 1.1), so this is missing tools rather than missing groundwork.

### Proposed Solution

Four tools in a new `tools/news.py`:

| Tool | Endpoint |
|---|---|
| `list_redmine_news` | `GET /news.json`, optionally narrowed to one project |
| `get_redmine_news` | `GET /news/{id}.json?include=comments,attachments` |
| `manage_redmine_news` | `POST /projects/{id}/news.json`, `PUT /news/{id}.json` |
| `delete_redmine_news` | `DELETE /news/{id}.json` |

Writing needs Redmine 5.1 or newer; an older server answers with an error rather than appearing to work.

Following the conventions already in the tree:

- `delete` is its own tool, like `delete_redmine_issue` and `delete_file`, and refuses without `confirm_delete=True`, returning an impact preview (title, comment and attachment counts). Keeping it separate is also what lets a deployment offer announcements without offering their destruction, since the actions inside a `manage_X` tool cannot be exposed selectively.
- `manage_redmine_news` therefore carries only `create` and `update` — the shape `manage_time_entry` already has.
- `summary`, `description` and every comment's `content` go through `wrap_insecure_content`. `title` stays unwrapped, for the same reason an issue's `subject` is: it is short, label-shaped, and rendered as an identifier.
- Write actions go through `@action_dispatch`, so `REDMINE_MCP_READ_ONLY` is honoured without a per-tool check.

Two details of this API are worth handling inside the tools rather than passing on to the caller:

**1. The list endpoint is effectively flat.** There is a nested `/projects/:id/news.json`, but python-redmine's `News.query_filter` is `/news.json` with no placeholder, so `project_id` travels as a query parameter for both `news.filter(project_id=...)` and `project(...).news`. Redmine reads that parameter — but a filter Redmine does *not* read is not an error there, it answers 200 with the collection unnarrowed, which a caller cannot tell apart from a filter that matched everything. The result is therefore checked against what was asked for and refused with a named code if it comes back unfiltered, rather than handing over a plausible superset.

**2. Create answers 201 with no body.** `NewsManager` compensates by re-reading `news.filter(**params)[0]`, i.e. the newest visible news item — probably, but not certainly, the one just created. The read-back is verified against the title that was sent; when it cannot be confirmed, the tool returns `confirmed: False` with a `CREATE_UNCONFIRMED` code and the values it sent, instead of a neighbour's record with a plausible id. The same applies to update, which Redmine answers with 204.

### Use Case

- "What's new in project X?" — `list_redmine_news(project_id=...)`
- "Summarize what happened in X this month" — the issue activity plus the announcements, which is where the release and maintenance notes are
- "Read me the announcement about the 2.4 rollout, with the comments" — `get_redmine_news(news_id, include_comments=True)`
- "Post the maintenance window for Saturday to the ERP project" — `manage_redmine_news(action="create", ...)`
- A deployment that wants announcements read-only exposes only the two read tools; one that wants posting but never deleting exposes `manage_redmine_news` and leaves out `delete_redmine_news`.

I have this implemented on a branch and can open the PR once you agree on the shape — `docs/contributing.md` asks for the issue first, so here it is.

### Feature Category

New MCP Tool

### Priority

Important

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Contribution

andilem:
Closes #269. One commit on top of `develop`.

Four tools — `list_redmine_news`, `get_redmine_news`, `manage_redmine_news` (create, update), `delete_redmine_news` — plus `news` in `search_entire_redmine`'s `allowed_types`, so the two land together as you asked.

## Your five corrections

**1. Dropped the project-filter guard.** Gone, along with `PROJECT_FILTER_IGNORED`. A bad `project_id` now surfaces as `NOT_FOUND` with the project echoed back, since `ResourceNotFoundError` is what python-redmine raises for that 404. I could confirm the narrowing half independently on an Easy Redmine instance before your comment arrived: filtering on a project whose news are *not* the globally newest returned only that project's items, so three servers agree.

**2. 204, not 201.** Fixed in the module docstring, the `CREATE_UNCONFIRMED` message, the comment on the `ResourceSetIndexError` branch, the tool reference and the changelog.

**3. Narrowed the read-back reasoning.** You are right that `self.params` carries the posted `project_id`, so the re-read is scoped to the target project. The docstrings now say the race is "someone else posting to the same project in the same instant" rather than "the newest news anywhere". `CREATE_UNCONFIRMED` stays.

**4. Scopes — one disagreement, please overrule me if I have it wrong.** The writes map to `manage_news` as you asked, and all four tools are in `TOOL_SCOPES` with a `ToolKind` each, the delete as `WRITE_DESTRUCTIVE_IDEMPOTENT`.

But I kept the reads on `view_news` rather than `view_project`, and I think `view_news` does exist. The evidence is in this repo rather than in my reading of the module: `tests/fixtures/redmine_6_permissions.txt` lists it at line 81, next to `comment_news` and `manage_news`, and that file's own header says where it came from —

```
# Redmine 6.1.1 stable -- Redmine::AccessControl.permissions.map(&:name)
# Generated from sandbox: docker exec redmine rails runner
```

— which is the same 6.1.1 sandbox you ran the endpoints against. So this is your generator's output, not my guess. One line settles it either way:

```
docker exec redmine rails runner 'puts Redmine::AccessControl.permissions.map(&:name).grep(/news/)'
```

Both names pass the fixture guard, so nothing is red either way. The reason I did not just follow you: `view_project` advertises a coarser scope than the endpoint needs, so a token holding `view_project` without `view_news` would pass our check and then be refused by Redmine — the scope map would be less accurate than Redmine's own gate. If the one-liner comes back without `view_news`, say so and I will switch; it is two lines.

**5. Both error paths named.** `NEWS_MODULE_DISABLED` for the 403, with a hint pointing at `get_project_modules` and a note that Redmine checks the module before permissions so an administrator is refused too. `NEWS_WRITE_UNSUPPORTED` for the pre-5.1 404 — and since a 404 on `POST /projects/{id}/news.json` is ambiguous between "no project" and "no route", the create path spends one extra request on the error path: if the project still reads back, the route is what is missing. Otherwise the plain `NOT_FOUND` stands.

## Counts

45 → 49, 58 → 62, 59 → 63, found by grepping the numbers rather than the files: README (3 places plus the category list), `docs/contributing.md` (2, including the module count 16 → 17), `docs/roadmap.md`, and `pages/index.html` (10 places, including the `<meta>` description and the JSON-LD FAQ block, which each carry their own copy). Nothing matching `45 core`, `45 tools`, `58 with`, `58 in total` or `maximum of 59` is left.

## Verification

2362 passed, 1 failed, `flake8` and `black` clean. The one failure is `test_ssl_integration.py::test_ssl_cert_file_exists`, which wants `tests/fixtures/ssl/ca-cert.pem` from `generate-test-certs.sh` and is unrelated.

`tests/test_news.py` has 30 tests. Beyond the happy paths they pin the three named failures, that a failed read-back after a successful write still reports success, that a create missing `title`, `description` or `project_id` is refused with the field named before anything reaches Redmine, that an empty `summary` on update clears the field rather than being dropped, and that reads keep working in read-only mode while all three writes are blocked.

Two existing search tests needed adjusting. `test_unknown_resource_type_skipped` used **news** as its example of a type outside `allowed_types`, which stops being true with this change, so it now uses `projects` — a type Redmine searches and this server still does not offer. The other two assert the default resource set and gained the third entry.

Live-tested against an Easy Redmine 4.4.2 instance (Redmine core new enough for the writes): list global and per project, get with both includes, create, update. The 403 path showed up on its own there — a project with the news module off — which is what prompted `NEWS_MODULE_DISABLED` before I read your comment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
+- News tools: `list_redmine_news`, `get_redmine_news`,
+  `manage_redmine_news` (create, update) and `delete_redmine_news`. News was
+  the last core Redmine resource with no coverage, and there was no
+  workaround either -- `search_entire_redmine` could not reach it. Reading
+  works on any Redmine; writing needs 5.1, where the REST API gained it.
+  Deleting is a separate tool, like `delete_redmine_issue` and `delete_file`,
+  so a deployment restricting its tools can offer announcements without
+  offering their destruction. Comments come back read-only, because Redmine
+  has no endpoint for adding one.
+  Three failure shapes get names instead of being passed on: a create whose
+  204-with-no-body read-back does not match the title that was sent reports
+  `CREATE_UNCONFIRMED` rather than a neighbour's record; a 403 on a write
+  reports `NEWS_MODULE_DISABLED`, since Redmine checks the news module before
+  any permission and refuses an administrator too; and a 404 on create where
+  the project still reads back reports `NEWS_WRITE_UNSUPPORTED`, because
+  `News.redmine_version` is `(1, 1, 0)` for the whole resource and
+  python-redmine raises no version error of its own.
+  ([#269](https://github.com/jztan/redmine-mcp-server/issues/269))
+- `search_entire_redmine` now searches news alongside issues and wiki pages,
+  which the module docstring already claimed.
 - Issue serializers pass through top-level keys the standard Redmine API does
   not define, under `unmapped_fields`. Distributions and plugins add their own
   keys to the issue JSON (Easy Redmine sends `easy_sprint` and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `manage_redmine_news` (create, update) and `delete_redmine_news`. News was
   the last core Redmine resource with no coverage, and there was no
   workaround either -- `search_entire_redmine` could not reach it. Reading
-  works on any Redmine; writing needs 5.1, where the REST API gained it.
+  works on any Redmine; writing needs 4.1, where the REST API gained it.
   Deleting is a separate tool, like `delete_redmine_issue` and `delete_file`,
   so a deployment restricting its tools can offer announcements without
   offering their destruction. Comments come back read-only, because Redmine
@@ -35,7 +35,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the plain error; and a 404 on create where the project still reads back
   reports `NEWS_WRITE_UNSUPPORTED`, because `News.redmine_version` is
   `(1, 1, 0)` for the whole resource and python-redmine raises no version
-  error of its own.
+  error of its own. That message names the endpoint rather than a core
+  version, since distributions vary in what they expose.
   ([#269](https://github.com/jztan/redmine-mcp-server/issues/269))
 - `search_entire_redmine` now searches news alongside issues and wiki pages,
   which the module docstring already claimed.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,12 +28,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   has no endpoint for adding one.
   Three failure shapes get names instead of being passed on: a create whose
   204-with-no-body read-back does not match the title that was sent reports
-  `CREATE_UNCONFIRMED` rather than a neighbour's record; a 403 on a write
-  reports `NEWS_MODULE_DISABLED`, since Redmine checks the news module before
-  any permission and refuses an administrator too; and a 404 on create where
-  the project still reads back reports `NEWS_WRITE_UNSUPPORTED`, because
-  `News.redmine_version` is `(1, 1, 0)` for the whole resource and
-  python-redmine raises no version error of its own.
+  `CREATE_UNCONFIRMED` rather than a neighbour's record; a 403 reports
+  `NEWS_MODULE_DISABLED` when reading the project's modules back shows the
+  news module really is off -- Redmine checks it before any permission and
+  refuses an administrator too -- while an ordinary permission denial keeps
+  the plain error; and a 404 on create where the project still reads back
+  reports `NEWS_WRITE_UNSUPPORTED`, because `News.redmine_version` is
+  `(1, 1, 0)` for the whole resource and python-redmine raises no version
+  error of its own.
   ([#269](https://github.com/jztan/redmine-mcp-server/issues/269))
 - `search_entire_redmine` now searches news alongside issues and wiki pages,
   which the module docstring already claimed.
@@ -98,6 +100,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   this class of bug has now shipped three times.
 
 ### Contributors
+- @andilem proposed and implemented the news tools
+  ([#269](https://github.com/jztan/redmine-mcp-server/issues/269))
 - @gino8080 reported that the issue serializers drop the top-level keys
   distributions and plugins add
   ([#263](https://github.com/jztan/redmine-mcp-server/issues/263)) and
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ A Model Context Protocol (MCP) server that connects AI assistants to Redmine. It
 
 ## Features
 
-- **45 MCP tools on a stock Redmine, 58 with the RedmineUP and DMSF plugins** (plus 1 operator tool gated by `REDMINE_MCP_EXPOSE_ADMIN_TOOLS=true`): Issues, projects, time tracking, wiki, Gantt, file operations, membership management, products, contacts and deals (CRM), DMSF documents, and more
+- **49 MCP tools on a stock Redmine, 62 with the RedmineUP and DMSF plugins** (plus 1 operator tool gated by `REDMINE_MCP_EXPOSE_ADMIN_TOOLS=true`): Issues, projects, news, time tracking, wiki, Gantt, file operations, membership management, products, contacts and deals (CRM), DMSF documents, and more
 - **Interactive Kanban Board**: `show_triage_board` renders a live, drag-and-drop issue board right in the chat via the MCP Apps extension
 - **Flexible Authentication**: API key, username/password, or OAuth2 per-user tokens
 - **Prompt Injection Protection**: User-controlled content wrapped in boundary tags for safe LLM consumption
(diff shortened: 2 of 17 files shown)


jztan (changes requested):
@andilem You are right about `view_news` and I was wrong. It exists on both sandboxes and it is not public, so keep your mapping.

Two 403 fixes before I merge.

1. The classification never runs on the reads. `list_redmine_news(project_id=1)` against a project with the news module switched off returns "Access denied. Your Redmine user lacks the required permission" instead of `NEWS_MODULE_DISABLED`. Run it in `list_redmine_news` and `get_redmine_news` as well, and rename the helper.

2. A plain permission denial reports a disabled module. With `manage_news` off the role and the module left enabled, create still came back `NEWS_MODULE_DISABLED`, telling the operator to switch on something that was already on. Disambiguate it the way you disambiguate the 404: read the project back and claim the module only when it really is off. The changelog and the two docstrings need the same correction.

Please also add yourself to `### Contributors` under Unreleased.

Everything else passed on 6.1.1 and 7.0.0. Push those two and I will merge.

andilem:
Both fixed in `d8833cd`.

**1. The classification now runs on all five paths**, so the helper is no longer write-only and is renamed `_classify_news_failure`. The 404-means-no-route branch stays write-only, since reading news works on every version.

**2. The 403 is disambiguated the way the 404 already was.** The project's modules are read back, and `NEWS_MODULE_DISABLED` is claimed only when `news` really is absent. An ordinary denial keeps the plain permission error, and an unreadable module list claims nothing rather than guessing. Docstrings, tool reference and changelog carry the corrected reasoning.

One limitation falls out of that, and I pinned it with a test rather than papering over it: `get_redmine_news` knows only the news id, so when reading the item is *itself* refused there is no project to check the module on, and nothing is claimed. In practice I expect Redmine to answer 404 there anyway — a news item in a module-less project is not visible, so `News.visible` filters it before any 403 — which the existing `NOT_FOUND` path already covers. The realistic read-allowed-write-refused case is `update`, and that one does resolve the project and does name the module. `delete` resolves it from the item it already read for the impact preview. If you saw a 403 rather than a 404 from a single-item read on either sandbox, tell me and I will find another way to reach the project.

While in there: the `enabled_modules` normalization moved to `_serialization` as `_enabled_module_names`, so `get_project_modules` and the news classifier share one copy instead of two.

`@andilem` is in `### Contributors` under Unreleased.

I also merged `develop` in, because #274 landed on the same `### Added` block and the changelog would have conflicted on your merge button. Upstream's entry kept its position, mine sit below it.

2378 passed, 1 failed, `flake8` and `black` clean. The failure is the unrelated `test_ssl_cert_file_exists`, which wants the certs from `generate-test-certs.sh`.

Thanks for the `view_news` check — good to have it settled from the sandboxes rather than from the fixture alone.

jztan (changes requested):
@andilem Thanks for turning both of those round so quickly, and for pinning the `get_redmine_news` limitation with a test rather than hiding it. Both fixes hold up on the sandboxes.

One correction before I merge, and it predates this round: news writing is Redmine 4.1, not 5.1. Redmine's REST API index says "CRUD operations other than index were added in 4.1" for News, and `accept_api_auth` in `news_controller.rb` picks up create, update and destroy at tag 4.1.0. What 5.1 added is `:create, :new` on the top-level `resources :news`, the global news form, which python-redmine never calls since it posts to `/projects/{id}/news.json`. Could you fix the version in the two README lines, the tool reference, the three docstrings and the changelog?

I would keep the guard itself, just stop it naming a core version. Neither of us has actually seen that 404: your writes went through on Easy Redmine, and core has routed those three endpoints since 4.0. Have it say the server does not expose the news write endpoint, with 4.1 as the floor and a note that distributions vary. That is the case it really protects.

Also curious what your instance reports under Administration > Information, since Easy Redmine's numbering does not track core.

And a small one: please put the `_custom_fields_to_list` import in `projects.py` back where it was, the move only breaks the alphabetical order.

andilem:
You are right, and thank you for chasing it down to the tag — 4.1 it is. I checked
your two pieces of evidence against the source before changing anything:

- `accept_api_auth` in `news_controller.rb` is `:index` alone at 4.0.0 and
  `:index, :show, :create, :update, :destroy` at 4.1.0.
- `config/routes.rb` has carried the project-scoped `resources :news, :except =>
  [:show, :edit, :update, :destroy]` (so index, new, create) plus the top-level
  `:update, :destroy` since 4.0.0. The 5.1.0 line adds `:create, :new` to the
  top-level `resources :news` — the global form, which python-redmine never
  reaches, since `NewsManager` posts to `/projects/{id}/news.json`.

So the endpoints were routed before 4.1 and API-authorized from 4.1, which makes
4.1 the floor for anything going through this server. Fixed in the two README
lines, the tool reference, the three docstrings and the changelog.

The guard stays and stops naming a core version:

```
error: This Redmine does not expose the news write endpoint over the REST API.
hint:  The project exists and is readable, so the missing piece is the
       endpoint, not the project. Core Redmine has exposed it since 4.1, but
       distributions vary. Reading news is unaffected.
```

Your framing is the better one: neither of us has seen that 404, our writes went
through on Easy Redmine, and core has routed those endpoints since 4.0 — so no
core version explains it, and a stripped or proxied distribution is the case the
branch actually protects. The test now pins exactly that split: the diagnosis has
to name the endpoint, and 4.1 has to stay in the hint and out of the error.

`_custom_fields_to_list` is back in alphabetical order — the move was accidental,
`_enabled_module_names` just landed where the sort said it belonged and I did not
re-sort the rest.

On your question about Administration > Information: it reports no core Redmine
version at all. The page opens with

```
Easy Redmine:
  Platform version   15.6.4
  Build version      20260310052050
  Full version       15.6.4
Environment:
  Rails version      7.2.3
```

and goes straight on to Ruby, Rails, SCM and the plugin list. Rails 7.2 is what
Redmine 6.0 requires, so the lineage is 6.x, but that is my inference from the
Rails version and not something the instance states. Worth knowing too: the 4.4.2
that turns up in Easy Redmine's own API description is their Swagger plugin's
version, not a Redmine one.

Which is the argument for the guard in one line: on this distribution not even an
administrator can read a core version off the instance, so an error message has
no business inferring one.

jztan:
@andilem Merged, thanks. Good catch that the 4.4.2 was the Swagger plugin's version, that explains where the whole 5.1 story started. And your point that the instance reports no core version at all is a better argument for keeping the guard than the one I gave you.

PR #276, opened 2026-09-08
Document uploads where a client can actually read it

### LINKED ISSUE #275: `uploads` is documented in the reference but not in the docstring, so an agent attaching a file reaches for `file_path` and cannot succeed on an HTTP deployment
### Bug Description

`create_redmine_issue` and `update_redmine_issue` accept an `uploads` list, and
`_build_upload_descriptors` resolves each entry through `_resolve_upload_content`, which
takes `content_base64`, `source_url` or `file_path`. `docs/tool-reference.md` explains
all of that correctly. The docstrings do not: neither tool has an `Args:` entry for
`uploads`, so the parameter reaches `tools/list` as

```json
"uploads": {"anyOf": [{"items": {"type": "object"}, "type": "array"}, {"type": "null"}]}
```

with no description and no key names. A client reads the schema, not the repository, so
an agent holding image bytes has no way to learn that `content_base64` is accepted
there. `manage_redmine_wiki_page` documents its own `uploads` in its docstring, which is
what makes the gap on the issue tools easy to miss.

What the agent finds instead is `upload_file`, whose docstring does describe all three
sources. Two things then go wrong:

1. `upload_file` writes to the project's *Files* section, not to the issue. It is the
   wrong tool for "attach this to the ticket", but it is the only one whose *docstring*
   mentions attaching a file at all.
2. Its `file_path` is resolved on the **server's** filesystem. On an HTTP deployment the
   server is a container on another host, so the caller's working directory does not
   exist there. No value of `REDMINE_MCP_UPLOAD_FILE_ROOTS` can make a client-side path
   resolvable — not a wider root, not any root.

The error message points at exactly that impossible fix:

```
file_path is outside the allowed upload roots. Allowed roots default to
ATTACHMENTS_DIR; widen them with the REDMINE_MCP_UPLOAD_FILE_ROOTS environment variable.
```

It names the operator knob and stays silent about the two sources that need no
configuration at all, so the caller is steered away from what works toward what cannot.

This is a documentation defect, not a broken write path. All three sources work: on a
Docker deployment behind HTTP I attached files through `uploads` with `content_base64`
on create, `content_base64` on update, and `source_url` on update. Each landed on the
issue byte for byte, and an `attachment:"name.png"` reference in the description
rendered inline. Only `file_path` failed, correctly.

It surfaced in practice: an agent working for a colleague reported back that "the
connector only accepts files from a shared directory (`REDMINE_MCP_UPLOAD_FILE_ROOTS`),
which my working directory is not in", and asked its user to attach the image by
drag-and-drop instead. It had the bytes in memory the whole time.

### Steps to Reproduce

1. Run the server over HTTP (e.g. the Docker compose setup) with a client on a different
   machine.
2. Inspect the tool schema:
   ```
   tools/list → update_redmine_issue → inputSchema.properties.uploads
   ```
   It is an untyped array of objects with no description; nothing names
   `content_base64`, `source_url` or `file_path`.
3. Give an agent a local image and ask it to attach the image to an issue.
4. Watch it reach for the only documented route and fail:
   ```python
   update_redmine_issue(
       issue_id=123,
       fields={"notes": "mockup"},
       uploads=[{"filename": "mockup.png", "file_path": "/home/me/mockup.png"}],
   )
   # {"error": "uploads[0]: file_path is outside the allowed upload roots. ..."}
   ```
5. Follow the error's advice and add `/home/me` to `REDMINE_MCP_UPLOAD_FILE_ROOTS`. It
   still fails: that directory is on the client, not in the container.

### Expected Behavior

The working route is discoverable from the tool surface alone:

- `uploads` is documented in the docstrings of `create_redmine_issue` and
  `update_redmine_issue` — the same treatment `manage_redmine_wiki_page` already gives
  it — naming the per-entry keys (`filename`, `content_base64`, `source_url`,
  `file_path`, `content_type`, `description`), saying exactly one content source is
  given, and saying `content_base64` is the one to use when the caller holds the bytes.
- The upload-roots error says the path is read **on the server**, and names
  `content_base64` and `source_url` as the way to send a file that lives on the caller's
  machine, so a reader on a remote deployment stops looking for a configuration fix that
  does not exist.
- The reference doc's `file_path` bullets say the same, since a `file_path` that is
  simply *absolute* reads as usable from anywhere.

### Actual Behavior

- `uploads` carries no description in the schema; both issue docstrings pass over it in
  silence, while `docs/tool-reference.md` documents it in full — in the one place the
  model never sees.
- The only docstring that mentions attaching a file is `upload_file`'s, and that tool
  targets the project's Files section rather than an issue.
- The roots error names only `REDMINE_MCP_UPLOAD_FILE_ROOTS`, so callers on an HTTP
  deployment are sent to an operator change that cannot work, and never learn that
  `content_base64` would have succeeded with no configuration whatsoever.

### Version

2.14.0

### Deployment Method

Docker

### MCP Client

Claude Desktop

### Redmine Version

_No response_

### Relevant Logs

```shell

```

### Additional Context

_No response_

andilem:
Fixes #275.

docs/tool-reference.md has described the three content sources of `uploads` all along, but a client reads the tool schema, not the repository. Neither issue tool documented the parameter in its docstring, so it reached tools/list as a bare array of objects with no description and no key names.

An agent holding a file therefore cannot discover that `content_base64` is accepted on the issue tools. What it does find is `upload_file`, whose docstring documents all three sources -- and whose `file_path` is read on *this server's* filesystem. Over HTTP that is a different machine from the caller's, so no value of REDMINE_MCP_UPLOAD_FILE_ROOTS can make a caller's path resolvable, while the error named only that variable. Reported from a deployment where an agent gave up and asked its user to attach the file by hand, holding the bytes the whole time.

- `create_redmine_issue` and `update_redmine_issue` grow an Args section covering `uploads`, the way manage_redmine_wiki_page already documents its own.
- The upload-roots error says the path is the server's and names content_base64 and source_url, which need no roots at all.
- The reference doc's file_path bullets say the same.

Documentation only; every write path already worked. Verified against a Docker deployment behind HTTP: content_base64 on create, content_base64 on update and source_url on update each attached the file byte for byte.


jztan (changes requested):
Thanks @andilem. The schema-vs-repo framing is right, and the `Args:` block does come
through: `uploads` now carries a real description in `tools/list`. Seven things before
I merge.

1. `filename` is not required. Only `content_base64` needs it; `source_url` derives it
   from the URL or `Content-Disposition`, and `file_path` from the basename.
   `docs/tool-reference.md` says exactly that in this same PR, so the two now
   contradict each other. Drop the `(required)` from both docstrings.
2. The 50 MiB cap is not specific to `content_base64`. `_resolve_local_file` and the
   download path both check `_FILE_UPLOAD_MAX_SIZE_BYTES`, so it belongs in the header
   the way `manage_redmine_wiki_page` has it: "Maximum 10 items, 50 MiB each".
3. Naming only `REDMINE_MCP_UPLOAD_FILE_ROOTS` makes `file_path` look unusable until an
   operator sets something. Mention the `ATTACHMENTS_DIR` default alongside it.
4. "a path on the caller's machine cannot be read here" holds only when the server is on
   another host. Running locally they are the same machine, `file_path` works, and it
   costs no tokens. Make the caveat conditional on that split.
5. `upload_file` is the signpost your issue says the agent lands on, and its docstring is
   still silent about `uploads`. One line there pointing at the issue tools for "attach
   this to a ticket" closes the route that misled your colleague's agent.
6. `manage_redmine_wiki_page`'s docstring should follow the new issue text, both the
   server-filesystem caveat and the advice on `content_base64`. Right now it tells the
   caller to prefer the other two sources.
7. CHANGELOG: move the entry to Fixed, link #275 and this PR, and add a Contributors
   entry for yourself.

andilem:
Thanks — all seven are in `f969c8d`.

1. `(required)` is gone from `filename` on both tools. It now says: required with
   `content_base64`, derived from the URL or `Content-Disposition` for `source_url` and
   from the basename for `file_path`. That was my error, and the reference doc in this
   same PR was the thing contradicting it.
2. Moved to the header as "Maximum 10 items, 50 MiB each". You are right that every
   source is measured: `_resolve_local_file` after the read, the base64 branch after
   decoding, and `_ssrf._download_file_url` twice, against the declared length and again
   while streaming. The module comment above `_FILE_UPLOAD_MAX_SIZE_BYTES` was stale in
   the same direction — it recommended `file_path` "for larger files" — so I corrected
   that too.
3. `ATTACHMENTS_DIR` is named alongside the roots variable everywhere `file_path`
   appears now.
4. Made conditional. The bullets now say the path is read where the server runs, so on a
   local deployment it reaches the caller's own files and costs no tokens; the gap opens
   only when server and caller are different hosts. That also fixes the ordering: I had
   `content_base64` first as if it were the default, when it is the fallback for the case
   where the cheaper two do not apply.
5. `upload_file` now opens with what it is — the project's document store — and points
   at `create_redmine_issue` / `update_redmine_issue` for an issue and
   `manage_redmine_wiki_page` for a page. Its own `file_path` bullet and Args entry
   carry the filesystem caveat, since that is the text the agent in #275 actually read.
6. `manage_redmine_wiki_page` follows. I kept its advice to prefer `file_path` or
   `source_url` — sending the file through the model is a real cost and worth steering
   away from — and added that `file_path` is read where the server runs, so a file on
   the caller's machine travels as `content_base64` unless the two are the same host.
7. Entry moved to Fixed with both links, and a Contributors entry added.
   `scripts/release.py` parses it: `collect_changelog_contributors` returns `andilem`
   in the Unreleased set.

Suite is green apart from `test_ssl_cert_file_exists`, which fails on my machine for
want of the fixture certificate and is untouched by this branch. `flake8` and `black`
clean.

jztan:
Merged, thanks @andilem. The part I liked was you catching the stale comment above
`_FILE_UPLOAD_MAX_SIZE_BYTES` on your own: it had been pointing people at `file_path`
for large files long before any of this. Reordering the bullets so `content_base64`
reads as the fallback is also better than what I asked for.

Issue #261, opened 2026-09-02
Per-user auth for Redmine instances without Doorkeeper: bind a browser-supplied API key to a self-issued OAuth token


andilem:
### Problem Statement

On a Redmine that cannot offer Doorkeeper, this server has no working multi-user story. The README is explicit about the requirement — "OAuth2 is the one hard requirement: it needs Redmine 6.1+ for Doorkeeper" — and every existing auth mode fails for a different reason once Doorkeeper is off the table:

- `legacy` shares one `REDMINE_API_KEY` across everyone. Every caller acts as that account, so `assigned_to_id="me"` is meaningless and Redmine's own permissions stop being a boundary. Fine for a single operator, wrong for a team.
- `legacy-per-user` is the right model but needs an `X-Redmine-API-Key` header per request. That header has to come from somewhere, and in a centrally distributed setup there is nowhere: a Claude Code plugin or a Claude Desktop connector ships one manifest to everybody, so the value is either identical for all users or an environment variable each person provisions on their own machine. The latter works but means a long-lived Redmine credential sitting in plaintext env on every workstation, and it breaks the moment the deployment is meant to be zero-setup.
- `oauth` and `oauth-proxy` both need Doorkeeper's `/oauth/authorize`, `/oauth/token` and introspection.

Easy Redmine is a concrete instance of this. On a current install (Easy Redmine API 4.4.2) `/oauth/authorize`, `/oauth/token` and `/admin/applications` all return 404 — the Doorkeeper engine is simply not mounted, and there is no admin switch to mount it. The same applies to any Redmine below 6.1 and to instances whose admins will not enable it.

The existing OAuth issues (#185, #188, #189, #220) all improve behaviour *given* Doorkeeper. I could not find one covering its absence, so this is meant to record the gap rather than duplicate a thread.

### Proposed Solution

A fourth auth mode in which the MCP server is its own authorization server and the only credential it ever sees is an API key the user pastes into a browser prompt.

The flow reuses what FastMCP already provides (`OAuthProvider` / `InMemoryOAuthProvider`), so the moving parts are small:

1. The client discovers the server as its own AS (RFC 9728 protected-resource metadata, RFC 8414 AS metadata, dynamic client registration, PKCE S256) — all of which FastMCP serves.
2. `/authorize` redirects to a login page served by the server itself.
[...]

PR #286, opened 2026-09-14
feat(auth): api-key-login provider and encrypted binding store (#261)

andilem:
First of the two PRs you asked for in #261, against the design in #265. This one is the provider, its store and the wiring. The browser page, the `_client.py` branch and the revocation middleware are PR 2, so after this merges the mode is selectable but not yet usable end to end.

## What is here

`ApiKeyLoginProvider` subclasses FastMCP's `OAuthProvider`, so `/register`, `/authorize`, `/token`, `/revoke` and the two well-known documents come from the framework with `issuer = REDMINE_MCP_BASE_URL`. What the subclass adds is the consent step: `authorize()` parks a transaction and returns `{base_url}/login?txn=…`, and `complete_login()` validates the pasted key against `/users/current.json`, applies the admin gate, binds the key and mints the code.

Storage mirrors what `OAuthProxy` builds internally: the operator secret goes through `derive_jwt_key` twice, the fingerprint of the result names the directory under `FASTMCP_HOME/api-key-login/`, and a `FileTreeStore` is wrapped in `FernetEncryptionWrapper` with `raise_on_decryption_error=False`, so rotating `REDMINE_MCP_JWT_SIGNING_KEY` signs everyone out instead of crashing the server. The store arrives as a constructor argument, so tests inject a `MemoryStore`.

Wiring: a lazy branch in `_select_auth_provider`, the mode added to `AUTHENTICATED_AUTH_MODES` so the well-known routes and `/health` are mounted, `/health` joined to the `legacy-per-user` probe, and `_client.py` no longer warning about missing credentials while noting that `REDMINE_API_KEY` is ignored. Legacy and the other OAuth modes do not import this module; I checked.

## Two properties with tests of their own

**Revocation never enumerates.** Access and refresh tokens hold a `binding_id` and dereference it on load, so deleting the binding invalidates every token pointing at it. That is what makes the store contract — `get`/`put`/`delete`/`ttl`, with `keys()` optional and not guaranteed through the encryption wrapper — sufficient.

**Single use rides on `delete()`.** The transaction, the code and the rotated refresh token are each claimed by deleting them and proceeding only on a `True` result. The one counter that cannot work that way, `attempts`, is approximate under concurrency and says so in the code and in its test. Rotated refresh tokens leave a tombstone, so presenting one again is recognised as reuse (RFC 6819 5.2.2.3) and burns the session.

## Three places I departed from the spec

**`BindingProtection` is in this PR, not PR 4.** The spec puts the token-derived implementation in the fourth PR but calls the interface a structural requirement, and adding it later would mean reworking what this PR introduces. Only `ServerSecretBindingProtection` ships here and it is a pass-through — the store's Fernet wrapper does the work. Say the word and I will pull it back out.

**`build_api_key_login()` lives in `_api_key_login.py`,** not `_env.py`, matching `build_oauth_proxy()` in `_oauth_proxy.py`: the factory sits next to what it builds and the import stays lazy.

**`fetch_redmine_identity()` lives with the provider,** not in the routes module the spec lists it under, because `complete_login()` is what calls it.

## Tests

45 unit tests against a `MemoryStore`, plus two that exercise the real encrypted file store, including that the API key is not recoverable from the directory and that a different secret gets its own directory rather than failing to decrypt.

Full suite on Windows: 2425 passed, 1 failed — `test_ssl_cert_file_exists`, which wants `tests/fixtures/ssl/ca-cert.pem` from `generate-test-certs.sh` and is unrelated to this change.


jztan (changes requested):
@andilem thanks for this, and for pulling `BindingProtection` in now. You're right that adding it in PR 4 would have meant reworking this one, so it stays, and the other two placements are fine. Some things to fix before I merge:

1. **`iss` breaks root deployments.** With `REDMINE_MCP_BASE_URL=https://host`, the metadata issuer is `https://host/` but the redirect carries `iss=https://host`, and the MCP SDK client rejects that (RFC 9207 is an exact string match). Please pass `iss=str(self.issuer_url)` as upstream does, and test the value against the metadata issuer for both a root and a subpath base URL.
2. **Raise OAuth errors instead of `ValueError`,** which currently comes out as a 500. Use `RegistrationError("invalid_redirect_uri")` in `register_client`, `AuthorizeError` in `authorize`, and `TokenError("invalid_grant")` in the exchange and refresh paths, including when the binding is gone before the code is exchanged.
3. **Re-check the redirect allowlist in `authorize`,** not only the exact match against the registration (decision 6). Otherwise tightening the allowlist never reaches clients that are already registered, and every authorize extends their TTL.
4. **Validate the key format in `complete_login`** with the `[A-Za-z0-9]{20,128}` rule from `_per_user.py` before it reaches httpx. A paste with a trailing space or newline raises `LocalProtocolError`, which is a `RequestError` with the whole key in its message, and it lands on the "Redmine unreachable" path.
5. **Make `fetch_redmine_identity` fail closed.** A 200 without an integer `id` or without `admin` has to count as a failed login, not as a non-admin user. 3xx, 5xx and non-JSON answers should raise one exception type that the route can map to 502.
6. **Make `browser_nonce` a required argument** so the provider can't skip the check, and compare CSRF and nonce as bytes, since `compare_digest` raises on a non-ASCII `str`.
7. **Keep the original grant on the rotated refresh token** (RFC 6749 §6). A narrowed refresh should narrow only the access token. Right now it narrows the session for good.
8. **Fail at startup when `REDMINE_API_KEY_LOGIN_SESSION_DAYS` is 0 or less,** and log the "ignored in this mode" warning for the variables the spec lists under Configuration.
9. **Key transactions and codes by hash, like tokens.** Both are bearer capabilities, and today they're plain filenames. Please also reword the module docstring: `FileTreeStore` never deletes expired files, so "no pruning job is needed" isn't true. I'll open a separate issue for the cleanup, since oauth-proxy has the same gap.
10. **Three tests still pass with their guard removed.** Put `admin` in `scopes_supported` for the admin test, advance the clock and assert that the client TTL grew, and decrypt in the same directory for the wrong-secret test. The rotation test should also assert that the old refresh token fails, and each fix above needs its own test.

I'll merge this and #287 together with no release in between, since on its own the mode is selectable but has no login page and no client branch.

andilem:
All ten done, each with a test. `788fa6b` here, and `fe013c4` on #287 follows the provider changes.

1. **`iss`** now comes from `self.issuer_url`, as upstream does. Parametrised test asserts it equals the metadata issuer for a root and a subpath base URL — the root case failed before, exactly as you said.
2. **OAuth errors** replace every `ValueError`: `RegistrationError("invalid_redirect_uri")`, `AuthorizeError`, and `TokenError("invalid_grant")` in both exchange paths, including a binding revoked between minting and exchanging. `invalid_scope` for a widening refresh.
3. **`authorize` re-checks the allowlist.** Test registers a client, tightens the allowlist, and asserts the next authorize is refused.
4. **Key format checked in `complete_login`** before httpx. The test asserts `fetch_redmine_identity` is never awaited and the attempt is still charged. The login route's own duplicate check is gone on #287, along with the helper it needed.
5. **`fetch_redmine_identity` fails closed.** A 200 lacking an integer `id` or the `admin` flag now raises instead of reading as a non-admin — that one would have walked an administrator straight past the gate. Transport failures, 3xx, 5xx and non-JSON all raise `RedmineUnavailable`, parametrised over six shapes, plus a test that the message never carries the key (`LocalProtocolError` puts it there).
6. **`browser_nonce` is required** and both secrets compare as bytes through a helper; a test feeds it a non-ASCII value, which is what used to raise.
7. **The rotated refresh token keeps the original grant.** Test narrows a refresh, asserts the access token narrowed, then widens back on the next refresh.
8. **`REDMINE_API_KEY_LOGIN_SESSION_DAYS <= 0` refuses to start,** and the variables from the other modes are named in a startup warning.
9. **Transactions and codes are keyed by `sha256`.** Tests assert the raw id is absent from the store and the hash is present. The docstring no longer claims no pruning is needed — it now says an expired record stops being served but the file stays.
10. **The three weak tests fail without their guard now:** `admin` is advertised in the admin-drop test, the clock moves before the client TTL is re-read, and the wrong-secret test decrypts in the *same* directory instead of reading an empty one. The rotation test also asserts the old refresh token stops loading.

One correction to my own #287 text while I was in there: the route strips whitespace around the pasted key before it travels, so the newline case never reaches `complete_login` through the browser. The provider test covers it directly, and a new route test pins the stripping, since selecting a key in Redmine picks up a trailing newline more often than not.

Merging both together with no release in between makes sense to me.

jztan:
@andilem thanks, all ten check out on my side. One thing is left, from item 3. The refusal is an `AuthorizeError` raised in `authorize`, and the SDK delivers that as a redirect to the URI that is no longer allowed (`https://evil.example/cb?error=invalid_request&...`). For a bad redirect URI, RFC 6749 4.1.2.1 wants an error page, not a redirect. Please move the check into `get_client` and drop any registered redirect URI the allowlist no longer matches, like upstream `OAuthProxy` does. The SDK's `validate_redirect_uri` then answers 400 before `authorize` runs, and the check in `authorize` can go. Please add a test that the response is a 400 with no `Location` header.

andilem:
Right, and thank you — I had the refusal in the wrong place. `53e50e0`, merged into #287 as well.

`get_client` now drops registered redirect URIs the allowlist no longer matches, so the SDK's `validate_redirect_uri` answers before `authorize` runs and the check there is gone. The refusal is a 400 error page with no `Location`, instead of handing the error to the endpoint the allowlist had just judged untrustworthy.

Two tests: one that a tightened allowlist empties the stored list, one that drives `/authorize` over ASGI and asserts 400 with no `Location` header. I disabled the filtering to confirm both fail without it.

jztan:
@andilem thanks, the 400 checks out. One more before I merge: `test_client_record_ttl_is_refreshed_on_authorize` is flaky, and it failed #287's 3.14 job (`599.999828 > 599.999842`). Patching `_now` doesn't move `MemoryStore`'s clock, so both readings are "600 minus a few microseconds" and the comparison is a race. Please drop the TTL comparison and assert the rewrite directly: wrap `store.put` and check that `authorize` writes the client record to `clients` with `ttl=session_ttl`.

andilem:
Fixed in `1420c84`, merged into #287.

You were right about the cause: patching the provider's `_now` leaves `MemoryStore` on the real clock, so both readings were the same 600-minus-microseconds value and the comparison was a coin flip. The test now wraps `store.put` and asserts `authorize` writes the client record to `clients` with `ttl=session_ttl`, which is the property it was after and settles deterministically. I confirmed it fails with the refresh call removed, and ran it repeatedly.

PR #287, opened 2026-09-14
feat(auth): api-key-login browser page, bound client, revocation hook (#261)

andilem:
Second of the two PRs from #261, on top of #286. **Draft until #286 merges**, since it branches from it; I will rebase onto `develop` and mark it ready as soon as that lands. With this the mode works end to end.

## What is here

**The page.** Two routes on the FastMCP instance, so they live inside the mounted MCP app and answer at `<mount prefix>/login`. It names the client's registered name, the host the code will be sent to, and the Redmine the key travels to, because open registration plus a crafted authorize link is the vector this mode has to survive. A per-transaction cookie (`login_<sha256(txn)[:16]>`, HttpOnly, SameSite=Lax) binds the transaction to the browser that rendered it, and the per-transaction name lets two logins run in one browser without overwriting each other. The CSP carries no `form-action`, per your finding.

**The client branch sits before the bearer branch** in `_get_redmine_client()`, and that order is load-bearing rather than stylistic: the bearer branch fires on the mere presence of an access token without comparing the mode, and in this mode one is always present, so it would forward our own opaque token to Redmine.

**Revocation on 401 is middleware.** `BindingRevocationMiddleware.on_call_tool` reads the envelope after `call_next`, unwraps the `{"result": …}` form that `build_error_tool_result` produces, and revokes on `code == "AUTH_FAILED"`. The `AuthError` branch gained that code. 403 stays a `ForbiddenError` and an ordinary permission error, so a missing permission never costs a session.

`_scrub_error_message` now also redacts the key bound to the current request; it only knew about the environment's key, which in this mode is not the one that matters.

## Verification

**Unit:** 32 tests, the routes driven over ASGI so the cookie really round-trips. Two of them exercise the whole OAuth surface once FastMCP mounts it.

**Integration:** four live checks of what `/users/current.json` answers, since that is the one thing that can differ between versions. Run against stock **Redmine 6.1.1 and 7.0.0** in Docker; both pass. Configured through `REDMINE_URL` plus `REDMINE_API_KEY_LOGIN_TEST_KEY`, skipping with a reason when unset, the same shape as `test_oauth_integration.py`. I did not have your sandboxes, so these are fresh containers with the REST API switched on.

**Manual end to end**, twice. Against Redmine 6.1.1 with a key I created: full flow, `tools/list`, a real tool call, then the key reset in Redmine — the next call returned `AUTH_FAILED` and the one after that 401, so the "reset your key and every session dies" guarantee holds. A container restart afterwards kept the session, confirming the store on the volume. Then against a live **Easy Redmine (API 4.4.2)** with a real user's key: discovery, DCR, login, code exchange, 49 tools, `get_current_user` returning that user, `list_redmine_issues` returning their actual issues, an unauthenticated call refused with 401, and a working refresh.

Full suite on Windows: 2457 passed, 1 failed — the unrelated `test_ssl_cert_file_exists`.

## Two things for PR 3

**The admin gate locks out the operator.** The default refuses a Redmine administrator's key, which is correct — an admin key bypasses Redmine's own permission checks, so the mode's central promise would not hold for that user. But the person setting the server up is usually an administrator, so the very first login attempt fails with "Keys of Redmine administrators are not accepted here." That happened to us. Worth calling out in the docs next to `REDMINE_API_KEY_LOGIN_ALLOW_ADMIN`, with the better answer stated plainly: use a second, non-administrative account rather than opening the gate.

**`/revoke` rejects public clients, and it is not this code.** The MCP SDK's `RevocationRequest` declares `client_secret: str | None` with no default, so pydantic requires the field to be present, while `TokenRequest` defaults it to `None`. A client registered with `token_endpoint_auth_method: none` is never issued a secret, so omitting the field yields `400 {"error":"invalid_request","error_description":"client_secret: Field required"}` before `ClientAuthenticator` — whose docstring says public clients skip authentication — ever runs. Sending an empty `client_secret` works, and the session then really ends. The test says so at the call site. It affects any mode mounting the SDK handler, so it may deserve an issue upstream rather than a workaround here.


jztan (changes requested):
@andilem the live flow works on both sandboxes, at the root URL and under a subpath, and tools really do run as the bound user. Some things to fix before this leaves draft:

1. **`server.py` regression.** The new `if REDMINE_AUTH_MODE == "api-key-login":` took over the `else:` of the scope-enforcement check. Now `oauth` and `oauth-proxy` with enforcement on log "scope enforcement is DISABLED", and this mode with enforcement off logs nothing. Please put the block after the whole if/else, and test both warnings.
2. **The browser cookie doesn't stop replay.** `GET /login` sets the stored nonce on every render, so a second browser that opens the same URL after the victim gets the same cookie and can complete the transaction with its own key. Please set the cookie on the first render only. A later GET without the matching cookie gets the error page and no cookie, and a reload in the same browser still works.
3. **Revocation can fire on data.** `_envelope` falls back to the whole result and only compares `code`, so viewing a product whose code is `AUTH_FAILED` ends the viewer's session. Please require an `error` key as well.
4. **Revocation misses some 401s.** `get_private_notes` and `list_project_trackers` return `[envelope]`, which the unwrap doesn't look into. `import_time_entries` and the upload helper in `files.py` report the 401 with no code. Please unwrap a one-item list in the middleware and give those two paths `code: AUTH_FAILED` when the cause is an `AuthError`.
5. **A reset key doesn't end the session.** Redmine serves an unknown key as anonymous, not with a 401, on anything anonymous can read. On both sandboxes, after a key reset, `list_redmine_projects` returns the anonymous view and nothing is revoked. Only login-required endpoints answer 401. This is a gap in my spec, not your code, but I'd like the fix here. At every refresh-token exchange, revalidate the bound key with one `GET /users/current.json`. If that returns 401, or a different user id, revoke the binding and return `invalid_grant`. That caps the anonymous window at the access-token lifetime. I'll cover Redmine's "Authentication required" setting in the docs in PR 3, and I'll update the spec.
6. **A non-ASCII cookie gives a 500** in the route's `compare_digest`. Please use `_constant_time_equal` there as well.
7. **The integration test fails whenever it's configured.** `test_an_unreachable_redmine_raises_instead_of_denying` still expects `httpx.RequestError`, but #286 now raises `RedmineUnavailable`.
8. **The wiring isn't under test.** Removing the middleware registration in `server.py` or the `/login` import in `main.py` leaves the whole suite green. Please add one test through `main.app` that registers, logs in, calls a tool against a mocked Redmine that answers 401, and asserts the binding is gone. Also pin the HTML escaping of `client_name`, and check that the key stays out of the logs on the refused path, not only on success.

Before I merge the pair, please run your Easy Redmine end to end once more by hand on the final heads, and confirm you've read the final diff.

jztan:
@andilem I've updated the spec in #265, and it changes three items in my review, plus one addition:

- **Item 2:** set the binding cookie on the `/authorize` 302 instead of on the first render. Wrap the SDK authorize route through `get_routes`, and require the cookie on both GET and POST `/login`. A "first render" flag races on the store, and a one-shot render ticket gets used up by a `HEAD` from a link checker.
- **Item 4:** drop the `import_time_entries` and upload-helper part, since the refresh revalidation covers them. The `error` plus `code` match and the one-item list unwrap stay.
- **Item 5:** revalidate before claiming the old refresh token, with its own 5 s total timeout rather than `REDMINE_TIMEOUT`. Make `fetch_redmine_identity` tell 401 from 403. Only a 401, a changed user id or a newly admin user revokes; 403 and `RedmineUnavailable` keep the session.
- **New, in the provider:** `delete()` isn't atomic on `FileTreeStore`. Two concurrent deletes both returned `True` in 41 of 300 runs on macOS, and otherwise the loser raises `FileNotFoundError`. Please route the transaction, code and refresh-token claims through one `_claim` helper that holds a process-wide `asyncio.Lock` and treats `FileNotFoundError` as a lost claim, and test it on a real `FileTreeStore`.

The spec's testing list covers the rest.

andilem:
All eight plus the four amendments, in `9c4655d`.

1. **`server.py`.** You were right, and it was mine: the middleware block had taken over the `else` of the scope-enforcement check. Moved after the whole if/else, with the warning tested in both modes and both ways round. Both new tests fail with the regression put back.
2. **Cookie on the `/authorize` 302,** wrapped at the ASGI layer through `get_routes` — handing the SDK's endpoint to `Route(endpoint=...)` makes Starlette call it as a request/response function, which is a 500. GET requires the cookie now too. Tests: the redirect carries it, a second browser gets the error page and no cookie, a reload in the same browser still works, and an error redirect carries none.
3. **`error` is required alongside `code`,** so a record whose own `code` field reads `AUTH_FAILED` no longer ends the viewer's session.
4. **One-item list unwrapped;** a longer list is left alone. I dropped the `import_time_entries` and upload-helper part per your amendment.
5. **Revalidation at every refresh,** before the claim, with its own 5 s budget. `fetch_redmine_identity` now raises `RedmineForbidden` for a 403, so only a 401, a changed user id or a newly administrative user revoke; 403 and `RedmineUnavailable` keep the session. Six tests, one per outcome. Thank you for chasing this one down on the sandboxes — the anonymous fallback is not something I would have found from the code.
6. **`_constant_time_equal` in the route.** The test sends the value percent-encoded, which is how a browser would deliver it; sending it raw never leaves httpx.
7. **Integration test fixed** and re-run against Redmine 6.1.1: four passed.
8. **`test_api_key_login_wiring.py`** drives `main.app` in this mode from registration to a tool call Redmine answers with 401, and asserts the binding is gone. It imports the package under a `sys.modules` snapshot — reloading in place strips the shared FastMCP instance of its tools and took 80 unrelated tests down with it on the first attempt. Plus the client-name escaping and the key staying out of the logs on the refused path.

**The `_claim` helper** is in, under a process-wide lock, with `FileNotFoundError` treated as a lost claim. Tested on a real `FileTreeStore`: 25 rounds of six racing claims, exactly one winner each.

Full suite: 2504 passed, 1 failed — the missing `tests/fixtures/ssl/ca-cert.pem`, unrelated.

Still to do before this leaves draft: the Easy Redmine run by hand on the final heads, and reading the final diff end to end. I will report back on both rather than mark it ready quietly.

jztan:
@andilem thanks, most of it checks out, including the refresh revalidation live on the sandbox after a key reset. A few things before this leaves draft:

1. **The cookie can still be issued to the wrong browser.** The wrapper takes the txn from the Location header, and the SDK keeps the query string of a registered redirect URI. So a client registered with `http://localhost:1/login?txn=<victim's txn>` that triggers an authorize error gets the victim's cookie, and it then completed the victim's login with its own key in my test. Please don't read the Location at all. Have the wrapper set a per-request holder (`HOLDER.set([])`) before calling the app, have `authorize()` append its txn after the store put, and issue the cookie only for what's in the holder. A bare contextvar leaks into the next request on the same task. Add a test for the crafted-redirect case.
2. **The 5 s budget isn't total.** `httpx.Timeout(5.0)` is per phase, so a Redmine that sends a byte every 3 s kept revalidation going for 18 s. Please wrap the call in `asyncio.wait_for`.
3. **`revoke_binding` needs the same tolerance as `_claim`.** Two concurrent revokes of one binding raised `FileNotFoundError` in 229 of 300 runs on the real store, which is a 500 at `/token` when two refreshes follow a key reset. The same bare delete is in `drop_transaction`, in `_count_attempt`, and in the expiry path of `get_transaction`.
4. **`POST /login` without the cookie shouldn't drop the transaction.** It happens before the CSRF check, so anyone holding the URL can cancel the victim's pending login. Refuse without dropping, the same as GET.
5. **The GET error page should mention blocked cookies,** like the POST page does.
6. **`_CLAIM_LOCK` binds to the first event loop that contends it,** and a later contended use from another loop raises. Please keep one lock per running loop.
7. **The wiring test writes into the real FastMCP home.** `settings.home` is fixed at import, so setting `FASTMCP_HOME` has no effect; patch `settings.home` as the claim test does.
8. **Tests that don't catch their mutation yet:** a 403 at login (remove the `RedmineForbidden` catch and it's a 500), revalidation running before the claim, the three call sites using `_claim` rather than `store.delete`, the key staying out of the revalidation log lines, and `HEAD /login`.
9. **Minor:** inconclusive revalidation logs at INFO; the spec says warning.

The Easy Redmine run and your read of the final diff are still the last step, as you said.

andilem:
All nine, in `30384df`. Thank you for actually running the crafted redirect — that one was mine, and I would not have found it by reading.

1. **The cookie no longer comes from the response at all.** The wrapper seeds a per-request holder, `authorize()` appends its transaction after the store put, and the cookie is issued only for what the holder carries. Seeded per request, not left to default, for the leak you named. Test registers a client with `.../login?txn=<victim's>`, triggers an authorize error, and asserts no `Set-Cookie` and an untouched victim transaction.
2. **`asyncio.wait_for`** around the revalidation. Tested with an answer slower than the budget: it gives up quickly and keeps the session.
3. **All deletes go through `_forget`,** which now both serialises and tolerates. The loser's error turned out to be platform-specific — `FileNotFoundError` on your macOS, `PermissionError` here on Windows, and `PathSecurityError` when a peer removes the collection directory between the path check and the unlink. I measured 13 failures in 40 concurrent revokes before, none after.
4. **`POST /login` without the cookie refuses without dropping.** You are right that it precedes the CSRF check; the test now asserts the transaction survives.
5. **Same wording on both routes,** naming blocked cookies.
6. **One lock per running loop,** in a `WeakKeyDictionary`, with a test that two `asyncio.run` calls get different locks.
7. **The wiring test patches `settings.home`.** It had been writing into the real FastMCP home — good catch.
8. **Tests added for each mutation you listed:** the crafted redirect, a 403 at login (which is a 500 without the catch), revalidation ordering before the claim, all three claim sites, the key staying out of the revalidation log, and `HEAD /login`.
9. **Inconclusive revalidation logs at warning.**

Full suite: 2515 passed, 1 failed — the missing `tests/fixtures/ssl/ca-cert.pem`, unrelated.

Next from me, and the last of it: the Easy Redmine run by hand on this head, and my read of the final diff end to end. I will post both before taking it out of draft.

jztan:
@andilem thanks, all nine check out on my side, including the crafted redirect, the 5 s budget against a slow Redmine, and concurrent revokes on the real store. Two small things:

1. **`_forget` hides a failure that isn't a race.** With a store whose delete keeps raising `PermissionError`, `revoke_binding` returns `False`, the access token still loads, and the only trace is a DEBUG line. After an `OSError`, check `get()`: if the record is gone, it was a lost race and quiet is right; if it's still there, log an error and re-raise.
2. **Two of the fixes aren't pinned yet.** Removing the `wait_for` keeps the suite green, and so does removing the `OSError` tolerance in `_forget`. Please assert the refresh returns within the budget when the identity call is slower than it, and that a store raising `FileNotFoundError` on delete gives `False` rather than an exception.

andilem:
Both done, `835d139`.

1. **`_forget` confirms the loss now.** After an `OSError` it reads the record back: gone means a race and stays quiet, still there means an error log and the original exception. A `get()` that fails too counts as "still there" — a store answering badly is not a clean loss. You were right that the old shape was the worse failure of the two: `revoke_binding` answering `False` while the token keeps working is exactly the thing this mode must not do.

2. **Both are pinned, and I checked each fails without its fix** rather than assuming:
   - the budget test now uses a thirty-second identity call, so removing `wait_for` turns it into a timeout instead of a pass;
   - a stub store raising `FileNotFoundError` on delete gives `False`, and one raising `PermissionError` with the record still present raises. Removing the `OSError` branch fails both.

Full suite: 2517 passed, 1 failed — the missing SSL fixture, unrelated. (The commit message says 2516; I wrote it before the run finished.)

That is everything from the reviews. Left on my side, as agreed: the Easy Redmine run by hand on this head and my read of the final diff, then out of draft.

andilem:
Both of the last two are done, so this is out of draft.

**Read of the final diff.** Four things to fix, none behavioural, in `117a4fa`: import order in `_api_key_login.py` had drifted across the patch rounds, `_BindingCookieASGI` still took and stored a provider it stopped using when the cookie moved to the per-request holder, the module docstring still described single use as resting on `delete()` alone without the lock and tolerance `_forget` adds, and `complete_login`'s docstring still promised that a transport failure "propagates unchanged" when it raises `RedmineUnavailable`.

**Easy Redmine by hand, on `117a4fa`.** Fresh container against a live Easy Redmine (API 4.4.2), root base URL, admin gate open since that account is an administrator:

- the cookie rides on the `/authorize` 302;
- `iss` in the callback is `http://localhost:18090/` and the metadata issuer is the same string — the root case from your first review, byte for byte;
- code exchange 200, `tools/list` 49, `get_current_user` returns that user, `list_redmine_issues` returns their six open issues;
- no token: 401;
- refresh **with** the new revalidation: 200, and the new access token works;
- `/revoke` 200, the next call 401.

The revalidation passed silently, so no "inconclusive" line and no revocation. The whole container log contains no 40-character hex string; the only trace of the key is `bound key ...8f88`.

One thing worth recording, because it was accidental and reassuring: my first attempt handed the browser the `/login` URL directly while my script had swallowed the `/authorize` redirect. The browser therefore had no cookie and the page refused it — the replay you described in round 2, tripped by accident and behaving exactly as intended. A real client opens `/authorize`, so the browser receives the cookie itself.

Full suite on Windows: 2517 passed, 1 failed — the missing `tests/fixtures/ssl/ca-cert.pem`, unrelated to this work.

Ready when you are, and thank you for the review rounds. Four of the findings were real defects I would not have caught by reading: the `iss` mismatch, the anonymous fallback after a key reset, the crafted redirect stealing a cookie, and `_forget` staying quiet about a delete that never happened.

jztan:
@andilem thanks, both check out, and everything from the three rounds is verified on my side, including a clean merge with the current `develop`. I'm ready to merge #286 and #287 together as soon as you post the Easy Redmine run on the final head, as you mentioned.

andilem:
@jztan I think our comments crossed — the Easy Redmine run is already in the comment directly above yours (09:17 UTC, seven minutes before), and it was made on `117a4fa`, which is still the head of this branch. Nothing has been pushed since, so there is no newer head to test against.

The short version, so you do not have to scroll: fresh container against live Easy Redmine (API 4.4.2), root base URL — cookie on the `/authorize` 302; `iss` in the callback `http://localhost:18090/`, byte for byte the metadata issuer; code exchange 200; `tools/list` 49; `get_current_user` and `list_redmine_issues` both correct; no token 401; refresh with revalidation 200 and the rotated token works; `/revoke` 200 and the next call 401. No 40-character hex anywhere in the container log, only `bound key ...8f88`.

Happy to re-run it on whatever you merge if you would rather see it on the merge commit, but as things stand `117a4fa` is what I tested.

jztan:
@andilem merged, thanks. Your Easy Redmine runs covered what I couldn't test here. #261 stays open until the docs land.

PR #295, opened 2026-09-15
feat: register tools from out-of-tree extension modules

### LINKED ISSUE #294: Extension point for registering tools from an out-of-tree module
### Problem Statement

A deployment whose Redmine runs a plugin the server has no tools for has only one way to get those tools: edit the server's own tree.

The three tables a tool has to appear in are module-level constants populated by importing the server's own modules. `TOOL_KINDS` is read at decoration time by `_AnnotatingFastMCP.tool()`, so a tool without an entry silently loses its annotations. `TOOL_SCOPES` is consulted by `ScopeEnforcementMiddleware.on_call_tool`, which denies a tool with no entry. `PLUGIN_FLAGS` is what `apply_plugin_visibility` walks to enable or disable the `plugin:<family>` tag. `tools/__init__.py` imports a fixed list of modules, and nothing anywhere reads a module name from the environment.

So a private tool means a fork of this repository, rebased onto every release, and the forked tools sit outside the anti-drift tests that make those table entries mandatory in the first place: the fork takes on the maintenance cost of the pattern without the check that enforces it.

This is not an argument for shipping more plugin tools upstream. The server already carries flags for the popular plugins; in-house and site-specific plugins are a long tail no upstream can or should carry. What is missing is the seam.

### Proposed Solution

An environment variable naming importable modules, and a small public module for those modules to register through.

`REDMINE_MCP_EXTENSIONS` holds module names, comma or whitespace separated. `main.py` imports them in order, after `from . import tools` and before `apply_plugin_visibility(mcp)`, logging one line per registered family: its name, its enabled state and its tool count. An import failure or a registration conflict raises and fails startup, because a server that silently came up without half its tools is worse than one that did not come up.

A new module `redmine_mcp_server.extensions` exposes `register_extension(ExtensionSpec(...))`:

```python
from redmine_mcp_server.extensions import ExtensionSpec, ToolKind, is_true_env, register_extension

    family="acme_widgets",                      # becomes the tag plugin:acme_widgets
    enabled=lambda: is_true_env("REDMINE_ACME_WIDGETS_ENABLED"),
    tool_kinds={"manage_widget": ToolKind.WRITE_DESTRUCTIVE},
    tool_scopes={"manage_widget": {"list": frozenset(), "create": frozenset({"edit_issues"})}},
    advertised_read_scopes=(),
    advertised_write_scopes=(),
))
```

Semantics, all of them enforced rather than merely documented:

- `family` must not already be in `PLUGIN_FLAGS`; every key of `tool_kinds` and `tool_scopes` must be absent from `TOOL_KINDS` and `TOOL_SCOPES`; the two key sets must be equal. Any violation raises `RuntimeError` at import.
- Registration merges into the three existing tables. `advertised_scopes()` appends each registered extension's lists while its `enabled()` is true, write lists suppressed under read-only mode, in registration order and without duplicates. `configured_advertised_scopes()` is untouched and can still narrow the result.
- The module defines its tools after registering, decorated `@mcp.tool(tags={plugin_tag(family)})`, so annotations resolve at decoration time and visibility applies exactly as it does for the built-in plugin tools.
- `redmine_mcp_server.extensions` re-exports what a tool actually needs: the `mcp` instance, `plugin_tag`, the offloading decorators, `action_dispatch` and `ActionMode`, `ToolKind`, the per-request client getter and the configured Redmine URL, the shared error handler and read-only error, the insecure-content wrapper, and the small validation and environment helpers. An extension then depends on a stated surface rather than on underscore-prefixed file names that are free to move.

Nothing bypasses the existing machinery: an extension's tools land in the same tables, are enforced by the same middleware, are hidden by the same tag, and are subject to the rules the anti-drift tests already assert. That is also what keeps the change small. The contract exists; it simply has no public door.

One existing test has to become extension-aware: `tests/test_tool_annotations.py` asserts `len(TOOL_KINDS) == 63`. It should count the server's own entries, which is the same number in a test environment with no extensions loaded.

### Use Case

An organisation runs a Redmine with two in-house plugins of its own, each with a JSON API behind the usual Redmine permission checks. Its people already reach Redmine through this server with per-user OAuth, and want the same agent to reach those plugins' resources in the same conversation, under the same token, with the same permission model deciding what is allowed.

With the extension point, that is a private package of a few hundred lines installed beside the server in a derived image, plus one environment variable. The server itself stays exactly as released, which is what makes it upgradable: the extension is written against a stated surface, and a release that changes that surface fails the extension's own tests before anything is deployed.

Without it, the same organisation either forks this repository and carries the fork across every release, or runs a second connector. Both are worse for the maintainer too: the fork's tools are invisible upstream, and bug reports that come back are against code nobody upstream can read.

The mechanism is stated in terms of in-house Redmine plugins generally, not any particular one. Nothing about it is specific to the plugins that motivated it.

### Feature Category

Other

### Priority

Important

### Alternatives Considered

- **A fork.** Works, and is what a deployment does today. It costs a rebase on every release and puts the private tools outside the anti-drift tests.
- **A second MCP server beside this one.** Costs the user a second connector and a second consent, and re-implements the OAuth proxy, the per-request client and the scope middleware to reach the same Redmine as the same user.
- **Automatic discovery through a packaging entry point** instead of an environment variable. Rejected because it makes installing a package change the tool surface without the deployment saying so. Naming the modules keeps the loaded surface stated in one place, which is also what the startup log line reports.

### Additional Context

The implementation exists and is tested against the current release: the hook, the `extensions` module, the `main.py` import step, tests covering import order, table merging, name collisions, a misspelled module, advertised scopes for enabled and disabled extensions and the extension-aware size assertion, plus a README section and a CHANGELOG entry. It can be opened as a pull request as soon as the direction here is agreed. Naming and shape are open to change.

### Contribution

mmahmed:
## Summary

Fixes #294.

Adds an extension point so a deployment can register tools for a Redmine plugin this server does not ship tools for, without forking the server.

`REDMINE_MCP_EXTENSIONS` names importable Python modules, comma or whitespace separated. `main.py` imports them after `from . import tools` and before `apply_plugin_visibility(mcp)`. Each module registers a family through a new public module, `redmine_mcp_server.extensions`, which merges the family's entries into the three tables that already exist and appends its advertised scopes under its own feature flag.

The design point is that nothing bypasses the existing machinery. An extension's tools end up in `TOOL_KINDS`, `TOOL_SCOPES` and `PLUGIN_FLAGS` like every other tool, are enforced by the same scope middleware, are hidden by the same `plugin:<family>` tag, and are held to the same rules the anti-drift tests assert. The tables were already the contract; this gives them a public door.

With `REDMINE_MCP_EXTENSIONS` unset, which is the default, the import loop does nothing, the tables are exactly what they are today, and `advertised_scopes()` returns exactly what it returns today. No existing tool, scope, annotation or flag changes.

## Changes

- `redmine_mcp_server/extensions.py`, new. `ExtensionSpec` (keyword-only, frozen), `register_extension()`, and re-exports of the pieces a tool is written against: `mcp`, `plugin_tag`, the offloading decorators, `action_dispatch` and `ActionMode`, `ToolKind`, the per-request client getter and the configured Redmine URL, the shared error handler and read-only error, the insecure-content wrapper, and the validation and environment helpers. Extensions import from here and never from an underscore-prefixed module, so those stay free to move.
- `redmine_mcp_server/_extension_registry.py`, new. Holds the registered-extension list and the function that turns it into advertised scopes, importing nothing but `_env`, because `oauth_scopes.advertised_scopes()` runs while `server.py` is still executing its module body and `extensions.py` has to import `server` for `mcp`. Keeping the registry out of that cycle is what lets every OAuth-mode start succeed with the variable unset.
- `_env.get_extension_modules()`, new. Parses `REDMINE_MCP_EXTENSIONS` on comma or whitespace, ignoring empty entries.
- `main.py`. Imports each named module in order between the `tools` import and `apply_plugin_visibility(mcp)`, holds each import to the specs it registered (below), logs one line per registered family, `extension loaded: family=<family> enabled=<true|false> tools=<n>`, and, when anything registered, tells the auth provider the widened scope list.
- `server.refresh_advertised_scopes()`, new, and `RedmineAuthProvider.update_scopes_supported()`, new. The provider is built while `server.py` runs, before any extension module is imported, and both builders snapshot the scope list into it; without a refresh an extension's scopes reach `advertised_scopes()` but never the documents served at `/.well-known/`. The OAuth proxy already has `update_default_scopes()` for exactly this; the remote provider gets the equivalent, and its authorization-server metadata route now reads the same property the protected-resource route does. A stock server registers nothing and its provider is never touched.
- `oauth_scopes.advertised_scopes()`. Appends each registered extension's read list, and its write list unless the server is read-only, while that extension's `enabled()` is true. Registration order, duplicates removed. `configured_advertised_scopes()` is unchanged and still narrows the result.
- `tests/test_tool_annotations.py`. The two inventory assertions (the table size and the reviewed empty-scope set) subtract extension entries so they keep measuring this tree; the cross-checks still iterate the whole table, so an extension's tools must satisfy them too.
- README: the environment variable in the configuration table, and a section "Extensions: tools for in-house plugins" with a minimal module and the note that an extension's scopes must exist as Redmine permissions on the target instance and be ticked on the OAuth application there. `docs/tool-reference.md` gains a short "Extensions" subsection beside the allow list. `.env.example` and `.env.docker.example` document the variable. CHANGELOG entry under Unreleased.

Failure is deliberately loud, because the failure this replaces is a server that came up looking healthy with tools missing:

- A module that cannot be imported fails startup, naming the module and the underlying error.
- A `family` already in `PLUGIN_FLAGS` raises `RuntimeError`.
- A tool name already in `TOOL_KINDS` or `TOOL_SCOPES` raises `RuntimeError`.
- `tool_kinds` and `tool_scopes` with different key sets raise `RuntimeError`, which is the mistake the anti-drift tests exist to catch, moved to import time for code those tests cannot see.
- A malformed spec raises before any table is read: a non-string family, a non-callable `enabled`, a value in `tool_kinds` that is not a `ToolKind`, a `tool_scopes` value that is neither a `frozenset` of scopes nor a `dict` of actions to `frozenset`s, or an advertised list that is a bare string (which would expand per character into `scopes_supported`).
- After each import, the tools the module defined are compared with the specs it registered: a tool it defined but no spec declares, a tool a spec declares but it never defined, a pre-existing tool whose object was replaced (FastMCP's duplicate policy on this server is `warn`, which logs and replaces), or a tool missing the `plugin:<family>` tag each raise `RuntimeError` naming the module and the tools.
- All checks run before anything is merged, so a rejected registration leaves the tables exactly as it found them.

## Testing

```
uv run flake8 src/ --max-line-length=88        clean
uv run black --check src/ --line-length=88     clean
uv run pytest -m "not integration" tests/      2475 passed, 113 deselected
python tests/run_tests.py --all                  exit 0 (integration phase skips without a Redmine)
```

`tests/test_extensions.py`, new, covers: modules named in the variable are imported in order and after `tools`; a first import of `main` in a subprocess loads the configured extensions and logs the line; the served `/.well-known/oauth-protected-resource/mcp` document lists an enabled extension's scopes in both `oauth` and `oauth-proxy` modes, and a stock server's provider is not refreshed; every spec-shape rejection, the keyword-only construction, and that a rejected spec leaves every table untouched; each import-boundary mismatch (undeclared, undefined, replaced, untagged); `apply_plugin_visibility` hides a disabled family's tools and lists an enabled one's; a registered family's kinds, scopes and flag appear in the shared tables and no pre-existing entry changes; importing the same module twice registers once, while registering the same spec twice is rejected; a duplicate family, a duplicate tool name, and mismatched `tool_kinds` / `tool_scopes` key sets each raise; a module name that does not resolve fails startup rather than being skipped; an enabled extension's advertised scopes appear in `advertised_scopes()`, a disabled one's do not, a write list is suppressed in read-only mode, and duplicates are removed without reordering the built-in list; `get_extension_modules` parsing; the startup log line; and the existing annotation and scope anti-drift tests still hold with an extension loaded.

Also run: the server started locally with a fixture extension module named in `REDMINE_MCP_EXTENSIONS` and logged the line above; a Docker image built from this branch, with a small package installed on top and the variable set, started and logged one line per family.

## Checklist

- [x] Change works with both local (`uv run python -m redmine_mcp_server.main`) and Docker (`docker-compose up`) deployments

## Notes for review

Two choices worth disagreeing with explicitly, if you would rather they went another way:

- **Environment variable rather than an entry point.** An entry point would be discovered automatically, which is the thing to avoid: it would mean installing a package changes the tool surface. Naming the modules makes the deployment say, in one place, exactly what it loaded, and that is also what the startup log line reports.
- **Fail startup rather than skip a broken extension.** Skipping would leave a server answering `tools/list` without tools the operator believes are there. Refusing to start is the version an operator can diagnose.


jztan (changes requested):
@mmahmed thanks for this, and for opening #294 first. I'm taking the direction: naming the modules in an env var and failing startup loudly is the shape I want. A few changes before I merge:

1. Rebase onto develop. The conflicts in `main.py`, the README and the CHANGELOG are small.
2. Cover `api-key-login` mode, which landed after you branched. `ApiKeyLoginProvider` copies the scope list into `valid_scopes` and `_advertised` when it's built, and `refresh_advertised_scopes` falls through for it. An extension's scopes are never advertised or granted there, so every scoped extension tool is denied. Add a branch for it, and make the fallthrough raise when extensions registered scopes and the provider isn't one it recognizes.
3. If a tool accepts an action that its `tool_scopes` dict leaves out, that action runs with no scope check. In `_assert_import_matches_specs`, compare each dict's keys with the tool's `action` enum and refuse a mismatch, or an `action` that isn't a `Literal`.
4. When the tool registry can't be enumerated, raise instead of warning and skipping. Everything else here fails closed, and this should too.
5. Take `REGISTERED_EXTENSIONS` and `extension_advertised_scopes` out of `__all__` (`main.py` can import them from `_extension_registry`). Also say in the docs that `redmine_mcp_server.extensions` is provisional for now and may change in a minor release.
6. Add registered families to `plugin_flags` in `get_server_info`, so a client can see them and not only the log.
7. Move the README section to `docs/extensions.md`, keeping the env table row and a one-line link in the README. Cut the CHANGELOG entry to two or three lines.

mmahmed:
Thanks — all seven are in, rebased onto `develop` and pushed. The review response is a second commit so the diff against the original is readable; say the word if you would rather have it squashed.

**1. Rebase.** Done. The conflicts were the three you expected: the `_cleanup` import in `main.py`, the env-table row in the README, and the `Unreleased` block.

**2. `api-key-login`.** You were right, and it was worse than not advertising: `ApiKeyLoginProvider` copies the list into `valid_scopes`, `default_scopes` *and* `_advertised`, and `_granted_scopes` intersects the request with the last of those, so an extension scope was refused at `/register` and dropped at `/authorize` as well. `update_scopes_supported` on the provider now replaces all three, and `refresh_advertised_scopes` passes it `configured_advertised_scopes()`, the same source `build_api_key_login` uses, so `REDMINE_MCP_SCOPES` still narrows it. The fallthrough raises when the registry holds any advertised scope and the provider is none of the three; with nothing declared there is nothing to lose and it is left alone, which keeps a foreign provider working on a deployment that sets no extensions. Tests cover the served document in all three modes and the granting path directly.

**3. Actions without a scope entry.** `_assert_import_matches_specs` now reads the tool's `action` parameter out of the schema FastMCP built and requires the dict's keys to equal it exactly. A missing key, an extra key, `action: str`, `Optional[Literal[...]]` and a tool with no `action` at all each refuse to start. One wrinkle worth knowing: a single-value `Literal` renders as `const` rather than `enum`, so both are read, and there is a test pinning that.

**4. Unreadable registry.** Raises now. The allow list keeps treating the same registry as best-effort because it only spots typos with it; here it is the only thing tying a module's tools to its spec.

**5. `__all__` and the provisional note.** Both names are out; `main.py` and `tools/meta.py` import them from `_extension_registry`. The module docstring and `docs/extensions.md` both say the module is provisional and may change in a minor release, and the doc asks extensions to pin the server version.

**6. `plugin_flags`.** One key per registered family, named after the family, next to the built-in seven, reading the same `enabled()` the visibility pass does.

**7. Docs.** The README section is now `docs/extensions.md`; the README keeps the env-table row and a one-line pointer. The CHANGELOG entry is three lines. `tool-reference.md` keeps its short cross-reference, updated for the action rule, the three modes and `plugin_flags`, and `contributing.md` gains rows for `extensions.py` and `_extension_registry.py`.

`flake8` and `black --check` are clean and `tests/test_extensions.py` is 120 tests. The full suite passes here except seven `upload_from_url` tests that resolve `example.com`, which this machine's resolver will not answer; they are untouched by this branch.

jztan:
@mmahmed thanks, all seven are in. The api-key-login fix went further than I asked, covering `default_scopes` and the grant path as well.

One more before I merge: `register_extension` checks `family` against `PLUGIN_FLAGS`, but `agile` and `tags` aren't in that table, and `get_mcp_server_info` still lists them in `plugin_flags`. An extension with `family="agile"` loads without error and overwrites the real flag in that response. Please reject every `plugin_flags` key as a family name, with a test.

Two commits is fine, no need to squash.

mmahmed:
Good catch, and it was mine to make: the family check predates `plugin_flags` carrying extensions at all, so adding that merge quietly widened what a family name can collide with. Reproduced it before fixing — `family="agile"` registered without complaint and took the key.

Fixed in a third commit. The seven built-in keys and their accessors now live in one table in `_env.py`, `get_mcp_server_info` builds that half of its response from it rather than writing the calls out, and `register_extension` refuses a family that is one of them. Building the response from the same table is the part that matters: the two lists drifting apart is what opened this in the first place, so a family upstream adds later is reported and refused in the same commit, with nobody having to remember the second half.

The `PLUGIN_FLAGS` check stays exactly as it was, ahead of the new one, so the five names in both sets keep the message they already had. The new one says what the collision actually costs, which is a client reading the key to decide whether the built-in support is on and getting an unrelated extension's flag.

Tests: `agile` and `tags` each refused, which is the regression; `dmsf` still refused by the original check, pinned because adding a second check is exactly the edit that reorders the first out of the way; every key of the table refused, driven by iterating the table so a new built-in family is covered on arrival; and a refused registration leaving all four tables untouched. In `test_server_info.py` the built-in keys are pinned to that table in order, as a prefix so the assertion still holds on a deployment that has extensions loaded. `docs/extensions.md` now says outright that those seven names are reserved.

One judgement call worth flagging, since you may want it done differently: `PLUGIN_FLAGS` and the new table overlap but are not the same set, and `crm-shared` and `deal-products` are in the first only. A single table describing each family, with "hidden when off" and "reported in server info" as attributes, would collapse both checks into one. That reaches into built-in behaviour well past this fix, so I left the two checks. Happy to do it separately if you want it.

`flake8` and `black --check` clean; 2695 passed, with the same seven `upload_from_url` tests failing here because this machine's resolver will not answer for `example.com`. CI on the previous push was green on 3.10 through 3.14.

jztan:
@mmahmed merged, thanks. Building the reserved-name check and the server-info keys from one table is the right fix, since the two lists drifting apart is what caused the bug. Keeping the two checks separate is fine; I don't need the single family table.

Issue #289, opened 2026-09-15
Expired OAuth state files are never deleted from FASTMCP_HOME (oauth-proxy, api-key-login)


jztan:
### Bug Description

FastMCP's `FileTreeStore` stops serving a record once its TTL passes, but it never deletes the file. Its own docstring says so: "No built-in cleanup of expired entries". In `oauth-proxy` mode, every `POST /register` and `GET /authorize` writes a file under `FASTMCP_HOME/oauth-proxy/`, and both endpoints are unauthenticated, so on a public deployment the directory can grow without bound. Token rotation adds more files over time as well. The upcoming `api-key-login` mode (#261) uses the same store under `FASTMCP_HOME/api-key-login/` and has the same gap.

### Steps to Reproduce

1. Write a record to the encrypted `FileTreeStore` with `ttl=1`, the way `OAuthProxy` builds it.
2. Wait two seconds.
3. `get()` returns `None`, but the file is still on disk, and it stays there.

`OAuthProxy` writes client registrations and pending transactions through that same store, so the same thing happens on the endpoints above.

### Expected Behavior

Expired records are eventually removed from disk.

### Actual Behavior

They stay forever, and nothing prunes them.

### Version

v2.15.0

### Deployment Method

Both. Since #267 the Docker image keeps `FASTMCP_HOME` on the data volume, so the files survive rebuilds too.

### Proposed fix

- Each file stores `expires_at` in plaintext next to the encrypted value, so a sweep can find and delete expired files without the encryption key.
- Add that sweep to the existing background cleanup task (`_cleanup.py`, on `CLEANUP_INTERVAL_MINUTES`), for both `oauth-proxy/` and `api-key-login/`.
- Ask `py-key-value` upstream to implement its `cull()` protocol for `FileTreeStore`, so the sweep can go away later.

I'll pick this up after the api-key-login docs PR.

PR #298, opened 2026-09-15
docs(auth): drop the stale "no cleanup job yet" paragraph

andilem:
The api-key-login deployment section ended up with two **Expired records** paragraphs that say opposite things.

#293 added the accurate one, right under **Persistent store**: the server sweeps expired records every `CLEANUP_INTERVAL_MINUTES`, whatever `AUTO_CLEANUP_ENABLED` says. The older paragraph, four below it, survived from #292 and still tells operators the opposite — that expired files stay on disk "until a cleanup job exists", pointing at #289, which #293 closed.

Read in order, the section promises a sweep and then withdraws it. This removes the stale paragraph; the one from #293 stays.

Verified against `_cleanup.py` on `develop` rather than against the PR description: `start()` runs when `_attachment_cleanup_enabled() or _auth_state_sweep_applies()`, and the sweep is gated only on the mode, so the surviving paragraph is right on both counts.

The `#289` reference in `CHANGELOG.md` is the fix entry itself and is left alone.


jztan:
@andilem thanks for catching this, and for checking it against `_cleanup.py` rather than the #293 description. Two paragraphs saying opposite things would have tripped up anyone setting up api-key-login.

PR #299, opened 2026-09-15
feat(auth): token-derived binding protection for api-key-login (#265)

andilem:
PR 4 from the plan in #265: the second `BindingProtection` implementation. PRs 1–3 are in, and the seam from #286 was built for exactly this, so nothing in the provider changes.

## What it does

`REDMINE_API_KEY_LOGIN_BINDING_CRYPTO=token-derived` draws a random data key per binding, encrypts the payload under it with AES-256-GCM and keeps no copy. The data key is then stored once for each capability that may later need it — the authorization code, then every access and refresh token minted from it — each time encrypted under a key derived from that capability's own value (HKDF-SHA256, own salt and nonce per wrap).

The store therefore holds the **hash** of a token next to a data key encrypted under the **token**. The token exists only in the client's hands, so the volume plus `REDMINE_MCP_JWT_SIGNING_KEY` yields ciphertext. Rotation carries the data key to the new pair.

Default stays `server-secret`. This is the deployment requirement I mentioned in #261: a stolen backup must not yield working Redmine credentials even to someone who also has the operator secret.

## What it costs, and why that is in the docs in bold

Nothing server-side can read a stored key without a presented token. That permanently rules out a session admin UI, background revalidation of stored keys, and re-encryption when the signing key rotates. The revalidation at refresh is unaffected — it was built to run with the refresh token in hand, which is the part of the round-2 design that made this scheme possible at all.

Records written under one scheme cannot be read under the other, **in either direction**. Switching a running deployment ends every session and everyone logs in again. Nothing leaks and nothing is lost, but it wants a quiet moment; the docs say so and there are tests for both directions.

## One deviation from the spec, deliberate

The spec says no code outside the interface branches on the scheme. My first draft broke that in `build_api_key_login`, with an `isinstance` to pick the startup warning. Instead `BindingProtection` gained an abstract `storage_warning` property and the call site logs `protection.storage_warning`. That keeps the structural rule literally true, and it means a third scheme cannot be added without stating what it guarantees — there is a test that an implementation omitting it cannot be instantiated.

## Verification

43 new tests in `tests/test_api_key_login_crypto.py`, in three layers.

**The primitive.** Round-trips, the wrong secret, a flipped byte in each of the six stored fields, malformed records from disk answered with `None` rather than an exception (a raise there is a 500 at `/token`), a decrypted payload that is not an object, and both switchover directions.

**The flow.** The whole login → exchange → refresh → revoke path under the new scheme, a recording store asserting that nothing it was ever handed contains the key, the tokens, or the data key, and that the data key is carried by exactly the three capability records.

**The root-access simulation** — `test_the_operator_secret_does_not_open_a_real_store_on_disk`. Real `FileTreeStore`, real Fernet wrapper, full login. It then plays the attacker: peels the store's own encryption off every file with the operator secret, exactly as the server does, and looks for the key. It asserts the walk actually opened records first, so it cannot pass vacuously, and it ends by running the same walk under `server-secret` and asserting the key **is** there — the contrast is the claim.

I mutation-checked rather than trusting a green first run:

| Mutation | Caught by |
|---|---|
| data key written next to the wrapped one | `test_the_data_key_is_never_written_in_the_clear` |
| `unwrap` ignores the presented secret | 4 flow tests |
| plaintext `api_key` left in the sealed record | 5 tests, including the disk-level one |

Full suite on Windows: **2622 passed, 2 failed**, both pre-existing on a clean `develop` and both Windows-only — the missing `tests/fixtures/ssl/ca-cert.pem`, and `test_keeps_a_record_rewritten_while_the_sweep_read_it` from #293. The second one I looked into, since it is new: the test renames a replacement file over the record **while the sweep still holds the handle open**, which is a POSIX-only move. On Windows that `os.replace` raises `PermissionError` (WinError 5, confirmed directly), the sweep's `except OSError` skips the file, and the assertion can never hold. The product code is fine; I did not touch it here and can open it as its own issue if you want.

Docs: a new **Binding protection** section in `docs/api-key-login-auth.md` with the comparison table, how it works, what it does not protect against, and the switchover caveat; the security model's "what is stored" paragraph now points there instead of stating the default's guarantee as the only one; plus the README row, both env templates and a CHANGELOG entry.

Not yet done: the manual E2E against live Easy Redmine with the flag on. I will post it the way I did for #287 before you merge.


andilem:
The manual run I owed you, on `edd5205` against a live Easy Redmine (API 4.4.2), a fresh container with `REDMINE_API_KEY_LOGIN_BINDING_CRYPTO=token-derived`, root base URL, admin gate open since that account is an administrator.

**The flow.** `iss` in the callback is `http://localhost:18090/` and the metadata issuer is the same string; code exchange 200; `tools/list` 49; `get_current_user` returns that user; `list_redmine_issues` returns their open issues; no token 401; refresh with revalidation 200 and the rotated token still reaches the key; `/revoke` 200 and the next call 401.

**Reuse detection, by accident.** My first run put the old refresh token in after the rotation, which revoked the binding as designed — and left me with nothing to inspect, which is why there were two logins. Worth recording that it fired end to end under the new scheme: 401 `invalid_grant`, binding gone.

**The root-access view, which is the point of this PR.** With a live session in place, I ran the attacker's walk inside the container: peel the store's own Fernet layer off every record with `REDMINE_MCP_JWT_SIGNING_KEY`, exactly as the server does, then look.

- 18 files on disk, **no 40-character hex string in the raw bytes**;
- 11 records opened with the operator secret — so the walk demonstrably worked and is not passing vacuously;
- the `bindings` record opens to `{"crypto": "token-derived-v1", "ciphertext": "...", "nonce": "...", "session_expires_at": ...}` and nothing else;
- **no `api_key` field anywhere in the decrypted set, and zero 40-character hex strings**;
- each access and refresh token record sits under a filename that is the SHA-256 of its token and carries `wrapped_key` with its own `wrap_salt` and `wrap_nonce` — the data key, encrypted under a token the store does not contain.

That is the same assertion as `test_the_operator_secret_does_not_open_a_real_store_on_disk`, but against a real container holding a real Redmine key rather than a `tmp_path`.

**Logs.** No 40-character hex string anywhere in the container log; the only trace of the key is `bound key ...8f88`. The startup warning reads as intended:

> api-key-login mode active. Binding crypto is token-derived: stored Redmine API keys are encrypted under keys that exist only inside the tokens clients hold, so the store and REDMINE_MCP_JWT_SIGNING_KEY together do not open them. Nothing server-side can read a stored key without a presented token. The store is node-local, so a second replica will not see these sessions.

One aside from the same session, not part of this PR. `tests/test_auth_state_sweep.py::test_keeps_a_record_rewritten_while_the_sweep_read_it` from #293 cannot pass on Windows: it renames a replacement over the record while the sweep still holds the handle open, and there `os.replace` raises `PermissionError` (WinError 5, confirmed directly) rather than succeeding, so the sweep's `except OSError` skips the file and the file keeps its original content. The inode guard in `sweep_expired_auth_state` is fine; only the simulation is POSIX-only. Say the word and I will open it as its own issue with a skip or a Windows-shaped variant.

jztan (changes requested):
@andilem thanks for the Easy Redmine run. The `storage_warning` property is a better answer than the isinstance, and the disk-level contrast test is the proof I wanted.

One blocker: `_mint_code` writes the plaintext code into the code record (`"code": code`), right next to the data key wrapped under that code. A backup taken between login and code exchange, or while an abandoned code is still on disk, opens that user's API key with just the store and the signing key. I reproduced this against the real `FileTreeStore`. Your container walk ran after the code had been exchanged and its record deleted, so it couldn't see this. Please drop the field and have `load_authorization_code` return the `authorization_code` it receives; the api-key-login tests still pass with that change. Then add a pre-exchange snapshot to the disk test, and add the code to the `not in dump` check.

Two smaller things: cut the CHANGELOG entry to two or three lines, since the docs carry the detail, and add #299 to your Contributors line. And yes, please open the Windows issue for the #293 test. I'll take it.

andilem:
All four, in `0a5f09b`. The blocker was real and it was the whole guarantee, not an edge of it — thank you for catching it.

**1. The code beside the key it unwraps.** Confirmed before fixing, against a real `FileTreeStore` with a never-exchanged code on disk: opened the code record with the operator secret, read `code` out of it, derived the KEK from that, unwrapped the data key, opened the binding, recovered the API key. Full break with nothing but the volume and `REDMINE_MCP_JWT_SIGNING_KEY`.

You are right that my container walk could not have seen it — it ran after the exchange, when the record was already deleted. The five-minute window and any abandoned code were wide open.

`_mint_code` no longer writes the field; `load_authorization_code` returns the code it was handed. Nothing else read it — the record is found by `sha256(code)` and every caller presents the code — so the change is two lines plus a comment saying why the field must not come back.

**2. Tests.** Three, and I checked all three fail with the field restored:

- the disk walk is now taken **twice**, the first snapshot with an unexchanged code still on disk, and it asserts a code record was actually present so the snapshot cannot pass vacuously;
- the code joins the values asserted absent from everything the recording store is handed;
- `test_a_record_never_carries_the_secret_that_opens_it` — the general form rather than one field name: for every record carrying a `wrapped_key`, every string in that record is tried as the secret, and none may work.

I also replaced the field-name check in the disk test with the actual exploit: `_key_recovered_from` tries every string in every opened record against every wrapped key, and every recovered data key against every sealed record. The old `"api_key" not in dump` assertion passed while the store was handing the key over, so it was not testing what I thought it was.

**3. CHANGELOG** cut to six lines pointing at the docs, and #299 added to the Contributors line.

**4. Windows test:** #300, with the `PermissionError` reproduction. Yours.

Full suite: 2624 passed, 1 failed — the missing `tests/fixtures/ssl/ca-cert.pem` — plus #300 on this platform.

One thing I have not redone: the live Easy Redmine run was on `edd5205`. The change since is a removed field and a returned argument, and the disk test now covers the window the container walk could not reach, but say the word and I will run it again on the new head before you merge.

jztan:
@andilem merged, thanks. The test that tries every string in a record as the unwrap secret guards far more than the one field I pointed at. I merged develop into your branch first to resolve a CHANGELOG conflict with #295.

PR #304, opened 2026-09-17
docs: describe the uploads sources caller-first

### LINKED ISSUE #303: `uploads` still recommends `file_path` first, so an agent picks the one source that cannot work on an HTTP deployment
Follow-up to #275 / #276. That round made the `uploads` sources discoverable and gave the upload-roots error a way out. Both landed in 2.15.0 and both work. An agent still does not take the way out, because the description tells it to try `file_path` first.

## What the description says today

`create_redmine_issue`, `update_redmine_issue` and `manage_redmine_wiki_page` list the sources in this order, `file_path` first:

> - `file_path`: a path on **this server's** filesystem, inside `ATTACHMENTS_DIR` or a directory listed in `REDMINE_MCP_UPLOAD_FILE_ROOTS`. The bytes never pass through the caller, so **prefer it for a file that is already there** -- which, when the server runs on the caller's own machine, includes the caller's own files. Where the server is a different host, a caller-side path cannot be read here, whatever the roots are set to.

Three things work against the reader here:

1. It is **first**, and it is the only source carrying an explicit recommendation ("prefer it").
2. The condition that makes the recommendation true -- *the server running on the caller's own machine* -- trails as a subordinate clause, and **nothing in the tool schema says where the server runs**. A model cannot check it. `get_mcp_server_info` deliberately excludes internal hostnames and filesystem paths, so it cannot answer the question either.
3. `content_base64` comes **last** and is framed as the concession: "The way to send a file the caller holds and the server cannot reach ... at the price of moving the file through the conversation."

A model holding a local mockup reads "a file that is already there", recognises its own file, and picks the one source that on an HTTP deployment can never work.

## Field report

A user asked an agent to attach a mockup to an issue comment through the connector. The agent tried `file_path`, got the roots error, and reported back:

> The mockup itself I could not attach -- the connector only takes files from a directory released on the server side.

It then stopped. It was holding the bytes, and the error it had just read named `content_base64` explicitly.

## Verified against 2.15.0

Remote deployment, `server_version` 2.15.0, caller on a different host:

| call | result |
| --- | --- |
| `update_redmine_issue`, `uploads=[{file_path: "C:\...\mockup.png"}]` | rejected, and correctly **before** any write -- `_build_upload_descriptors` returns on the first failing entry, so the issue was untouched |
| `create_redmine_issue`, `uploads=[{content_base64, filename}]` | attachment created, 70 B, `image/png` |
| `update_redmine_issue`, `uploads=[{content_base64, filename}]` + `notes` | attachment created and placed on the new journal |

So the behaviour is right and only the description is wrong -- same shape as #275, one layer up.

The error text is already good, for what it is worth:

```
uploads[0]: file_path is outside the allowed upload roots. It is read on the
MCP server's own filesystem, not on the caller's: if the file is yours rather
than the server's, send it as content_base64, or as a source_url the server
can fetch, and no roots have to be configured at all. ...
```

An agent that reaches it has a way forward. The point of this issue is that it should not have to reach it: the first source it reads about should be the one that works wherever the server is deployed.

## Proposal

Order the sources caller-first in all four tools (`create_redmine_issue`, `update_redmine_issue`, `manage_redmine_wiki_page`, `upload_file`) and in `docs/tool-reference.md`:

1. `content_base64` -- how a file on the caller's own machine reaches the server. No configuration, works whatever the deployment.
2. `source_url` -- preferred over base64 whenever the file is already at a URL, since it spares the caller the bytes.
3. `file_path` -- for a file that genuinely lives on the server. Stated as a condition rather than a recommendation, and closing with: where the deployment is unknown, send `content_base64` rather than trying a local path first.

`upload_file` needs the same treatment: it currently leads with `source_url` and describes `content_base64` as "use this only when the caller already has the file content in memory", which reads more restrictively than it is.

Documentation only -- no write path changes. Happy to open the PR; tests pinning the ordering included so it cannot drift back.

andilem:
Closes #303. Follow-up to #275 / #276.

That round made `uploads` discoverable and gave the roots error a way out. Both shipped in 2.15.0 and both work. What did not change is the order the sources are described in: `file_path` first, and the only one carrying a recommendation ("prefer it for a file that is already there"). The condition that makes that true -- the server running on the caller's own machine -- trails as a subordinate clause, and nothing in the tool schema tells a model where the server runs. On the usual HTTP deployment the first source it reaches for is the one that can never work.

Reported from a deployment where an agent tried `file_path` for a local mockup, hit the roots error, and told its user the attachment was impossible -- while holding the bytes, and having just read an error that names `content_base64`.

## Change

Sources are described caller-first in `create_redmine_issue`, `update_redmine_issue`, `manage_redmine_wiki_page`, `upload_file` and `docs/tool-reference.md`:

1. `content_base64` -- how a file on the caller's own machine reaches the server. No configuration, works whatever the deployment.
2. `source_url` -- preferred over base64 whenever the file is already at a URL, since it spares the caller the bytes.
3. `file_path` -- a condition rather than a recommendation: for a file that genuinely lives on the server, closing with "where the deployment is not known, send `content_base64` rather than trying a local path first".

`upload_file` gets the same order; its `content_base64` no longer reads "use this only when the caller already has the file content in memory", which was more restrictive than the code.

Documentation only. No write path, no signature and no behaviour changes.

## Verification

Against a live 2.15.0 deployment with the caller on a different host:

| call | result |
| --- | --- |
| `update_redmine_issue` with `uploads=[{file_path: "C:\...\mockup.png"}]` | rejected, before any write -- the issue was untouched |
| `create_redmine_issue` with `uploads=[{content_base64, filename}]` | attachment created, 70 B, `image/png` |
| `update_redmine_issue` with `uploads=[{content_base64, filename}]` + `notes` | attachment created and placed on the new journal |

Two tests in `tests/test_upload_content_resolution.py` pin the ordering across all four tools and assert the old recommendation is gone, so it cannot drift back. Suite passes apart from two failures that are already present on `develop` and unrelated to this change (`test_ssl_cert_file_exists`, whose fixture is not in the repo, and `TestColdStartImport`, a local environment issue).


andilem:
Blind-tested against a deployment running this change. The reordering does what it was meant to do.

**Setup.** Fresh Claude Code session, no context from the work on this PR, started outside the repository so no project memory could leak the answer. A neutral ticket in a sandbox project, worded without any hint about attachments. A real 3232-byte PNG wireframe on the caller's machine, and the prompt the original field report came from:

> Hänge bitte das Mockup `C:\Users\alr\Downloads\login-dialog-mockup.png` als Kommentar an Ticket #43394 und schreib einen Satz dazu, was darauf zu sehen ist.
>
> *(Attach the mockup … to ticket #43394 as a comment, and write a sentence about what it shows.)*

**Result.** The first `uploads` call carried `content_base64`. No attempt at `file_path`, no roots error, no giving up. Before the change the same situation produced a `file_path` attempt, the roots error, and the agent telling its user the attachment was impossible.

So the ordering was the thing standing in the way, and this is worth merging on its own.

**One finding from the same run, which does not affect this PR but is worth knowing.** The attachment that landed was corrupt: 3208 bytes against the source's 3232, identical up to byte 907, diverging inside the IDAT stream, `IEND` missing. The transcript shows why. The agent ran `base64 -w0` into a file and `cat`-ed it; that tool output is complete and correct at 4312 characters. The string it then wrote into the tool call is 4280 -- 32 characters lost in a repetitive stretch of the payload. The model re-emitted the base64 by hand and lost its place.

That is not a defect in this change. If anything it sharpens it: `source_url` is not merely cheaper than `content_base64`, it is more reliable, and the base64 route is riskier the larger the file. I have opened #305 for it, proposing that the bytes take the same side channel the download direction already uses — a single-use upload ticket and a `POST` route mirroring `/files/{file_id}`, so a local file never has to be retyped by the model at all — with an optional per-item checksum as the fallback for payloads that still travel as base64.

jztan:
@andilem thanks for the careful write-up and for the tests that pin the order. I'm closing this in favor of #306. Your own evidence in #305 shows that base64 of a real file can arrive corrupted and still report success. If this ships alone, agents stop hitting a loud roots error and start attaching broken files without noticing, and I'd rather have the loud error. #306 already carries the reordering with `upload_id` in front, so nothing here is lost.

PR #306, opened 2026-09-17
feat(uploads): let a caller's file reach the server without the model

### LINKED ISSUE #305: A local file can only reach `uploads` by being retyped by the model, and it arrives corrupted
A file on the caller's machine has only one way into `uploads` on an HTTP deployment: `content_base64`. And there is no way to *pipe* a file into a tool argument — whatever ends up in that field is written by the model, character by character. For a real file that is a transcription task thousands of characters long, and it fails.

## Evidence

Observed on 2.15.0, attaching a 3232-byte PNG wireframe to an issue. The agent did everything right:

1. `base64 -w0 mockup.png > tmp` — correct
2. `cat tmp` — the tool output in the transcript is complete and correct, 4312 characters, correct head and tail
3. `update_redmine_issue` with `uploads=[{content_base64: ...}]` — **4280 characters**

32 characters lost. The first 1209 and the last 2459 match; the damage sits in between, exactly where the wireframe's flat colour areas compress into a near-repeating pattern:

```
correct: …MHQMHUmSoYO+hY+hIkgwdQwdDR5Jk6g6GDqSJEPH0DF0JEmGjo+joWP…
sent:    …MHQMHUmSoWPoYOhIkgwdQ8fQkSQZOhg6ho4kyDF0JEmGDaO…
```

Insertions, deletions and transpositions between offsets 1209 and 1334, plus two isolated substitutions further on. The model lost its place while copying.

What landed: 3208 bytes against the source's 3232, identical up to byte 907, diverging inside the IDAT stream, `IEND` missing. Redmine stored it, the API reported success, and the issue carries a broken image. The server had nothing to object to — valid base64, under `_FILE_UPLOAD_MAX_SIZE_BYTES`.

At 3 KB. A screenshot or a PDF is far worse, and the failure is silent every time.

## What does not solve it

- **Inside MCP.** The protocol has no client→server file transfer. `roots` tells a server which directories exist but grants no read access; sampling and elicitation move text. Every byte of a tool argument comes from the model.
- **`source_url` pointing back at the caller.** `_ssrf.py` rejects loopback, RFC1918 and link-local by design, and across hosts a caller-side URL does not route anyway.

## Proposal: let the bytes take the side channel that already exists

This server already moves attachment bytes past the model in the *other* direction. `mcp.custom_route("/files/{file_id}")` serves a downloaded attachment, guarded by an unguessable UUID with a TTL and swept by the cleanup manager; `get_redmine_attachment` hands the caller a URL, not the content. Uploading is the mirror image and can reuse the same machinery — `AttachmentFileManager`, the UUID directories, the expiry, the sweep.

1. **`create_upload_ticket()`**, an ordinary MCP tool and therefore authenticated. Returns `{upload_url, ticket, expires_at}`. The ticket is short, single-use, valid for minutes, and authorises nothing but one upload.
2. **`POST {upload_url}`** with `X-Upload-Ticket`, body `multipart/form-data` or `application/octet-stream`. The server stores the bytes in a UUID directory with a short TTL and answers `{"upload_id": ..., "size": ..., "sha256": ...}`. One `curl --data-binary @file` does it.
3. **A fourth source in `uploads`: `upload_id`**, resolved in `_resolve_upload_content` next to the existing three.

The model writes a path into a shell command and reads back a UUID. Nothing long passes through it, so nothing long can be miscopied. The security model is the one `/files/{id}` already uses — a capability string on a Starlette route — rather than bolting MCP auth onto custom routes, which today are not behind it.

Side benefit: a 5 MB attachment currently costs ~6.7 MB of base64 in the conversation even when it succeeds. This costs a path and a UUID.

## Secondary: narrow what `content_base64` is for, and make it fail loudly

`content_base64` should not be removed — that would break compatibility and strand clients that cannot run a shell. But its documented purpose is wrong, and it is the reason agents reach for it in the one situation it cannot serve. Today it reads:

> the way to send a file the caller holds and the server cannot reach

That is precisely the case that corrupts. What is actually left for it:

- **Clients with no shell.** The upload route needs someone to issue an HTTP request. Claude Code can; a browser-based connector cannot. There `content_base64` is the only source that works at all — but only for content the model itself produced (a CSV it assembled, an SVG it drew, a document it wrote). A browser client has no access to the user's files either way, so base64 does not rescue that case; it was never a file-transfer mechanism there.
- **Genuinely small payloads**, where one call beats a ticket, a `POST` and an `upload_id` for a few hundred bytes.
- **Deployments without the new route.**

So three changes alongside the transport:

1. **Restate the purpose**: for content the caller *generated* and that is small — not for a file the caller holds.
2. **Verify it.** An optional `sha256` (and/or `size_bytes`) per `uploads` item, checked after decoding and refused on mismatch in `_build_upload_descriptors` before anything is uploaded. 64 characters survive being copied in a way 4312 do not, and any corruption changes them. The description should tell an agent to take the checksum from the same shell invocation that encoded the file. In practice this wants to be the norm for `content_base64`, not an afterthought: without it, every base64 upload is taken on trust.
3. **Cap it**, with an error pointing at the upload route. Where the ceiling belongs is genuinely unknown — it depends on the model and on how repetitive the payload is, which is why a 3 KB wireframe of flat colour failed while plenty of larger files presumably have not. Erring low, in the low single-digit KB, and phrasing it as what has been seen to hold rather than as a guarantee, still beats a coin flip: a deterministic refusal that names the alternative is a better outcome than a silently broken attachment.

Note that (2) is a detector, not a fix. It turns a corrupt attachment into a clear error; the agent still has to retry the same unreliable copying. The transport above is what removes the failure mode.

Both are backwards compatible: without the new keys and the new route, nothing changes.

Happy to open the PR if the shape sounds right.

---

*Edited after filing: the first version led with the checksum as the fix and did not mention the transport at all. The checksum is the fallback; moving the bytes off the model's output path is the fix.*

andilem:
Closes #305.

A file on the caller's machine had exactly one way into `uploads` on an HTTP deployment: `content_base64`. There is no way to pipe a file into a tool argument, so whatever goes into that field is written out by the model, character by character — and a payload of a few thousand characters does not survive that reliably. The case in #305: `cat` put a correct 4312-character payload in the transcript, the tool call carried 4280, and the attachment that landed diverged from the source inside the IDAT stream with no `IEND` chunk. Valid base64, under the size cap, so the server stored it and reported success.

## The change

The bytes now take the side channel this server already uses in the other direction. `/files/{file_id}` hands out a downloaded attachment against an unguessable id; `POST /uploads/{upload_id}` accepts one against a single-use ticket.

```bash
# 1. reserve a slot  (MCP, authenticated)
create_upload_ticket(filename="mockup.png")
# -> {upload_url, upload_id, ticket, expires_at, max_bytes}

# 2. send the bytes  (one request, from the caller's shell)
curl -sS -H "X-Upload-Ticket: $TICKET" --data-binary @mockup.png "$UPLOAD_URL"
# -> {upload_id, size, sha256}

# 3. attach it
update_redmine_issue(issue_id=42, fields={...}, uploads=[{"upload_id": "..."}])
```

`upload_id` is a fourth content source on `create_redmine_issue`, `update_redmine_issue`, `manage_redmine_wiki_page` and `upload_file`. The model writes a path into a shell command and reads a UUID back; nothing long passes through it, so nothing long can be miscopied.

**Reuses what is there.** Staged files are ordinary UUID directories under `ATTACHMENTS_DIR` carrying the metadata shape the download path already writes, so `AttachmentFileManager` expires and sweeps them with no change. A failed attach can be retried without resending the file.

**Security.** The route is guarded by the ticket, not by the MCP session's auth — which does not reach these Starlette routes today — matching how `/files/{file_id}` guards downloads. A ticket is single-use, minutes long (`REDMINE_MCP_UPLOAD_TICKET_MINUTES`, default 15), size-bounded (`REDMINE_MCP_UPLOAD_MAX_BYTES`, default 50 MiB), stored only as a SHA-256, and authorises nothing but that one upload. A wrong ticket and an unknown id answer identically, so the endpoint cannot be used to probe which ids exist. `filename` is reduced to a basename before it becomes a path segment. `create_upload_ticket` is mapped scope-free because it touches no Redmine resource; whatever the staged file is later attached to is scoped where that attach happens.

## Alongside it, for the base64 route that remains

- **Optional `sha256` / `size_bytes` per item**, verified once the content is resolved and refused before anything reaches Redmine. 64 characters survive copying where 4312 do not, so a mangled payload fails loudly instead of being attached. Accepted on every source, and on `upload_file` directly.
- **`REDMINE_MCP_CONTENT_BASE64_MAX_BYTES`** caps that one source. **Left unset it is the ordinary upload limit, so nothing that works today stops working.** #305 floated a low default; I deliberately did not apply one, since it would break deployments that are attaching larger files successfully right now. That call is yours — say the word and I will set a default.
- **The sources are documented caller-first**, and `content_base64` is described as what it is actually good for — small content the caller *generated* — rather than as the way to send "a file the caller holds", which is the one case it cannot do.

## Note on overlap with #304

The reordering of the existing three bullets is also in #304. Whichever merges second rebases to nothing. If you would rather keep them apart, I can drop the reordering from this PR and rebase it onto #304.

## Tests

`tests/test_upload_ticket.py`, 29 cases over the real ASGI app: byte-for-byte round trip on a raw body and on multipart, wrong ticket, reused ticket, expired ticket, oversize, empty body, unknown id, ticket never stored in the clear, filename path traversal, `upload_id` as a content source including the explicit-filename and never-sent cases, the integrity check (including a reconstruction of the exact corruption from #305), and the base64 cap.

Full suite passes apart from eight `TestColdStartImport` cases that fail identically on a clean `develop` in my environment — a local asyncio import in the subprocess, unrelated to this change.

Two bugs the tests turned up while writing this, both fixed here: `os.rename` refuses an existing target on Windows, so rewriting a record needs `os.replace`; and keeping the used ticket's hash is what lets a retry hear "already used" rather than the blanket "unknown".


jztan (dismissed):
@andilem this is the right fix for #305, and reusing the `/files` machinery keeps it small. A few things before I merge:

1. `filename=".."` escapes `ATTACHMENTS_DIR`. The POST writes `<ATTACHMENTS_DIR>.tmp` into the parent directory, returns 500, and nothing ever sweeps the file. `"."` leaves `<uuid>.tmp` at the top of the dir. Please reject both in `_safe_filename` (fall back to `upload_<id>`) and add a test for each.
2. `GET /files/{upload_id}` serves a staged upload without the ticket. The upload_id is part of the upload URL, so it lands in proxy logs. Make `serve_attachment` return 404 for `kind == "upload"`, with a test.
3. Drop multipart. `request.form()` reads the whole body to disk before the size check runs, so the cap doesn't hold on that path. Raw body only; `curl --data-binary` covers it.
4. The wiki `uploads` section in `docs/tool-reference.md` still leads with `file_path` and says "Prefer `file_path`". Bring over the reorder from #304.
5. `create_upload_ticket` is a new core tool, so the counts need bumping. README says 49 core / 62 / 63, and `contributing.md` lists `files.py` as 4 tools.
6. Keep the base64 cap unset with no default, as you have it.
7. I'm closing #304, so drop the "overlaps with #304" line from the CHANGELOG.

andilem:
All seven addressed in 5961fb2. Thanks — 1 and 2 were real holes.

**1. `filename=".."` / `"."`** — confirmed and fixed. `os.path.basename("..")` is `".."`, so the staged path resolved to the *parent* of `ATTACHMENTS_DIR` and the temp file landed outside the tree the cleanup manager walks. Both names now fall back to `upload_<id>`. I also added `_staged_target`, which re-checks that the resolved path's parent is still the upload's own directory before it reaches an `open()` — one check felt thin for a path built from caller input. Two tests, one per name, asserting the staged path stays inside the uuid directory.

**2. `GET /files/{upload_id}`** — fixed. `serve_attachment` returns 404 when the record's `kind` is `upload`. The test posts a staged file and then asks for it by id, asserting both the 404 and that the bytes are not in the response body. Your reasoning is the part I had missed: the two ids do not carry the same weight, because an upload_id travels in a URL and a download id never does.

**3. Multipart** — dropped. You are right that `request.form()` buffers before anything can check the size, so the cap did not hold there. Raw body only now. I replaced the test that asserted multipart works with one that pins what happens instead: the MIME envelope is stored as the bytes it is, bounded by the cap, rather than silently unwrapped. The route docstring, the tool docstring, the tool reference and the CHANGELOG all say so.

**4.** Wiki `uploads` block reordered caller-first; the "Prefer `file_path`" sentence is gone, and `file_path`'s own bullet no longer claims a caller-side path might work.

**5.** 49 → 50 core, 62 → 63, 63 → 64 maximum, File Operations 4 → 5 with `create_upload_ticket` listed, and `files.py` 4 → 5 in `contributing.md`. `upload_file`'s one-line description now names `upload_id` first.

**6.** Unchanged — `REDMINE_MCP_CONTENT_BASE64_MAX_BYTES` stays unset with no default.

**7.** The "overlaps with #304" note is out of the CHANGELOG; the reorder lives here now.

`flake8`, `black --check` and `pytest -m "not integration"` are clean locally, bar eight `TestColdStartImport` cases that fail the same way on a clean `develop` in my environment and pass on CI.

jztan:
@andilem merged, thanks. Adding `_staged_target` as a second check behind the filename fix was a good call, and the ticket route finally gives local files a way in that the model can't garble.

Issue #146, opened 2026-06-04
create_redmine_issue returns "Requested resource not found" on success, causing silent duplicate creation


SailorMan64:
### Bug Description

When calling create_redmine_issue, the tool returns an error response 
of "Requested resource not found" even though the issue was successfully 
created in Redmine. The caller has no way to distinguish a genuine 
failure from a successful creation, leading to unnecessary retries and 
risk of duplicate issues.

Note: This bug report was drafted by Claude AI (Anthropic) during an 
active development session. The observed behavior is real and 
reproducible. Root cause analysis is AI inference — take the technical 
speculation with appropriate skepticism.

### Steps to Reproduce

1. Call create_redmine_issue with a valid project_id, subject, 
   description, and fields payload (tracker_id, status_id)
2. Observe that the tool returns: {"error":"Requested resource not found."}
3. Retry the same call with identical payload
4. Receive the same error response again
5. Check Redmine — the issue was created on one of the attempts

### Expected Behavior

On successful issue creation, the tool should return the new issue ID 
and a success response. "Requested resource not found" should only be 
returned when the resource genuinely does not exist (404). The caller 
should be able to trust the response as an accurate reflection of what 
happened in Redmine.

### Actual Behavior

create_redmine_issue was called three times with identical payloads. 
Each call returned {"error":"Requested resource not found."}. 
After all three calls, only one issue was actually created in Redmine 
(ID: 6957). Issue IDs 6955 and 6956 do not exist — it is unclear 
whether Redmine assigned and rolled them back, or whether only the 
third call reached Redmine successfully. The created issue is correct 
and complete.

### Version

0.11.0

### Deployment Method

PyPI installation (uv/pip)

### MCP Client

Claude Desktop

### Relevant Logs

```shell
No server-side logs captured. Client-side error response on all 
three attempts:
{"error":"Requested resource not found."}

Issue 6957 exists in Redmine confirming successful creation.
Issues 6955 and 6956 do not exist.
```

### Additional Context

Self-hosted Python venv behind Apache reverse proxy over HTTPS. 
Systemd service running as www-data. MCP endpoint proxied via 
Apache Location block from https://[host]/mcp/[instance]/mcp to 
127.0.0.1:8001.

Claude AI (claude.ai) via custom MCP connector URL

- Project has issue_tracking module confirmed enabled
- API key is admin-level with full read/write access
- Auth mode: legacy (API key)
- All three attempts used identical payloads including tracker_id 
[...]

PR #308, opened 2026-09-18
feat(projects): add manage_redmine_project

### LINKED ISSUE #307: Add a manage_redmine_project tool: nothing can write a project record
## Problem Statement

There is no way to write a project record. Every project tool the server has is either a read (`list_redmine_projects`, `get_project_modules`, `list_project_trackers`, `list_project_issue_custom_fields`, `summarize_project_status`, `list_project_members`, `list_redmine_versions`) or a write to something *inside* a project (`manage_redmine_version`, `manage_project_member`, `manage_issue_category`). Nothing can set a project's `name`, `description`, `homepage`, `parent_id`, `is_public`, `inherit_members`, `enabled_module_names`, `tracker_ids`, `issue_custom_field_ids`, its default version or assignee, or its custom field values, and nothing can create a project or close and reopen one.

An agent can therefore report on a project in detail and change everything in it, but cannot fix a typo in its name, enable the wiki module on it, reparent it, or close it when the work finishes. The user has to leave the agent and open the Redmine settings page, which is exactly the handoff this server exists to remove.

Redmine has served all of this over the REST API for a long time. `ProjectsController` declares `accept_api_auth :index, :show, :create, :update, :destroy, :archive, :unarchive, :close, :reopen` (checked at tag 6.1.1), and python-redmine — already a dependency at `>=2.5.0,<3` — supports create, update and the `close`/`reopen` pair through `ProjectManager`. The gap is only on this side.

## Proposed Solution

A `manage_redmine_project` tool in `tools/projects.py`, shaped exactly like `manage_redmine_version` beside it: `@mcp.tool()` over `@action_dispatch`, one `@offloaded` handler per action.

Four actions, `create`, `update`, `close`, `reopen`, with these permission gates (from `lib/redmine/preparation.rb` at 6.1.1):

| Action | Endpoint | Redmine permission |
|---|---|---|
| `create` | `POST /projects.json` | `add_project` (`:require => :loggedin`), plus `add_subprojects` on the parent when `parent_id` is given |
| `update` | `PUT /projects/{id}.json` | `edit_project` |
| `close` | `PUT /projects/{id}/close.json` | `close_project` |
| `reopen` | `PUT /projects/{id}/reopen.json` | `close_project` |

Writable fields are Redmine's own `Project` safe attributes: `name`, `identifier` (create only), `description`, `homepage`, `is_public`, `parent_id`, `inherit_members`, `enabled_module_names`, `tracker_ids`, `issue_custom_field_ids`, `default_assigned_to_id`, `default_version_id`, `default_issue_query_id`, `custom_fields`. No invented parameter names.

`identifier` is accepted on `create` and rejected on `update`, because Redmine freezes it once the project exists (`Project#identifier_frozen?`) and python-redmine lists it in `Project._update_readonly`. Passing it to an update would otherwise be discarded silently and reported as success.

`PUT /projects/{id}.json` answers `204 No Content` (`render_api_ok`), so `update`, `close` and `reopen` read the project back and return it. That is necessary to return anything at all, and it also lets the caller see whether a permission-gated field actually took.

**Two actions Redmine exposes that this deliberately does not.**

`archive` and `unarchive` are `accept_api_auth`, but `ProjectsController` gates them on `require_admin`, not on a project permission. `oauth_scopes.py` never advertises `admin` — by design, since a token carrying it bypasses every per-permission check — so no token this server issues could reach them. Shipping them would add two actions that always fail.

`delete` is reachable (`DELETE /projects/{id}.json` needs only `delete_project`; the `require_sudo_mode :destroy` filter returns early for API requests), and it is left out on judgement rather than capability. It destroys every issue, wiki page, document and file in the project and in each of its subprojects, and advertising `delete_project` would put "Delete the project" on every user's consent screen for a capability nothing else here needs. Happy to add it if you would rather the tool be symmetric with `manage_redmine_version`.

**Scopes.** `WRITE_SCOPES` gains `add_project`, `add_subprojects`, `edit_project`, `close_project`, `select_project_publicity` and `select_project_modules`. `TOOL_SCOPES` requires only the four endpoint gates, one per action.

The last two are worth a word, because they are the reason the list is six and not four. They gate a *safe attribute* rather than an endpoint: `Project`'s safe-attribute blocks require `select_project_publicity` for `is_public` and `select_project_modules` for `enabled_module_names`. Redmine drops an attribute the caller may not set instead of refusing the request, so a token without them still edits every other field — which is why they are advertised but not required, matching the boundary already stated in `TOOL_SCOPES`' header comment. Not advertising them would ship two parameters that can never work for anyone in OAuth mode, since `Role#allowed_permissions` intersects the role with the token's scopes.

Annotation: `WRITE_DESTRUCTIVE`, the conservative union — `update` overwrites existing settings and `close` takes a project out of service, though no action deletes anything.

Deployments upgrading in OAuth mode need the new permissions ticked on their Redmine OAuth Application before the server restarts, or new authorizations fail with `invalid_scope`. Worth a CHANGELOG note; I will put one in the PR.

## Use Case

As a project manager working through an MCP client, I want to set up and maintain a project without leaving the conversation, so that the agent that plans the work can also make the structural change the plan implies.

Example workflow:

1. "Spin up a subproject under Space for the lunar programme, with issue tracking and the wiki on." → `manage_redmine_project(action="create", name="Lunar Programme", identifier="lunar-programme", parent_id=7, enabled_module_names=["issue_tracking", "wiki"])`
2. "Point it at the new brief." → `manage_redmine_project(action="update", project_id="lunar-programme", homepage="https://example.com/apollo")`
3. Work happens: issues, versions and members through the tools that already exist.
4. "That programme wrapped last week, close it." → `manage_redmine_project(action="close", project_id="lunar-programme")`

Today step 1 is impossible, step 2 is impossible, and step 4 is impossible. Only step 3 works.

A second case, less obvious and the one that prompted this: an agent auditing a set of projects can already find the ones whose modules or custom field bindings are wrong, list them, and explain the fix — and then has to hand every one back to a human to click through. Reading a misconfiguration you cannot correct is a strange place for a tool to stop.

## Feature Category

New MCP Tool

## Priority

Important

## Alternatives Considered

**Redmine-side configuration.** Nothing here is a permission or role problem — the endpoints are simply not wired up on this side, so no Redmine change substitutes.

**A generic `fields` bag**, the way `create_redmine_issue` takes one. Rejected: projects have a small, stable, documented set of safe attributes, and named parameters put them in the tool schema where the agent can see them. `manage_redmine_version` next door already uses named parameters, so this matches its neighbour.

**Splitting into `create_redmine_project` / `update_redmine_project` / `close_redmine_project`.** Rejected as running against the v2 consolidation (#100) that produced the `manage_X(action=...)` shape in the first place.

**Building it outside the server** through `REDMINE_MCP_EXTENSIONS` (#294). Rejected: that hook exists for tools over a Redmine plugin's own API. This is core Redmine with no plugin involved, so it belongs in core.

## Additional Context

Every Redmine claim above is checked against the released tag 6.1.1 rather than a branch:

- `app/controllers/projects_controller.rb:25-35` — the `authorize` / `authorize_global` / `require_admin` filters and the `accept_api_auth` list
- `lib/redmine/preparation.rb:38-43` — `add_project`, `edit_project`, `close_project`, `delete_project`, `select_project_publicity`, `select_project_modules`, and `:47` for `add_subprojects`
- `app/models/project.rb:825-873` — the safe-attribute blocks, including the two conditional ones
- `app/views/projects/show.api.rsb` — the response shape the serializer mirrors
- `lib/redmine/sudo_mode.rb:155-165` — `SudoRequestFilter#before` returning early for `api_request?`, which is why `delete` is a judgement call and not a blocker

python-redmine 2.5.0: `Project.query_update` / `query_create`, `Project._update_readonly` containing `identifier`, and `ProjectManager.__getattr__` serving `close` / `reopen` / `archive` / `unarchive` for Redmine >= 5.0.

I have the change implemented against `develop`, with 30 tests covering the four actions, the two refusals, read-only mode, the serializer's absent-key handling and the scope table. Happy to adjust the action set or the scope list before opening it.

## Contribution

mmahmed:
Closes #307.

Adds `manage_redmine_project`, the first tool that writes a project record. Everything the server could do to a project until now was either a read or a write to something inside it — versions, memberships, categories — so an agent could describe a project in full and change everything in it, but could not rename it, reparent it, enable a module on it, create one, or close one.

## What it does

`manage_redmine_project(action=...)` in `tools/projects.py`, shaped like `manage_redmine_version` beside it: `@mcp.tool()` over `@action_dispatch`, one `@offloaded` handler per action.

| Action | Endpoint | Permission |
|---|---|---|
| `create` | `POST /projects.json` | `add_project`, plus `add_subprojects` on the parent when `parent_id` is given |
| `update` | `PUT /projects/{id}.json` | `edit_project` |
| `close` | `PUT /projects/{id}/close.json` | `close_project` |
| `reopen` | `PUT /projects/{id}/reopen.json` | `close_project` |

Parameters are Redmine's own `Project` safe attributes and carry its names: `name`, `identifier`, `description`, `homepage`, `is_public`, `parent_id`, `inherit_members`, `enabled_module_names`, `tracker_ids`, `issue_custom_field_ids`, `default_assigned_to_id`, `default_version_id`, `default_issue_query_id`, `custom_fields`.

## Three decisions worth reviewing

**`identifier` is create-only.** Redmine freezes it once the project exists (`Project#identifier_frozen?`), and python-redmine lists it in `Project._update_readonly`. Passing it to an update would be dropped silently and reported as success, so `update` refuses it with an error that says why.

**Every write reads the project back.** `PUT /projects/{id}.json` and both status endpoints answer `204 No Content` (`render_api_ok`), so there is nothing to return without a follow-up `GET`. That also makes the response the honest record of what was stored, which matters for the two permission-gated fields below.

**Six new advertised scopes, four required ones.** `WRITE_SCOPES` gains `add_project`, `add_subprojects`, `edit_project`, `close_project`, `select_project_publicity` and `select_project_modules`; `TOOL_SCOPES` requires only the four endpoint gates, one per action.

The last two gate a *safe attribute* rather than an endpoint — `Project`'s safe-attribute blocks require `select_project_publicity` for `is_public` and `select_project_modules` for `enabled_module_names`. Redmine drops an attribute the caller may not set instead of refusing the request, so a token without them still edits every other field. They are advertised so a token *can* carry them and deliberately left out of `TOOL_SCOPES`, matching the boundary already stated in that table's header comment. Leaving them unadvertised would ship two parameters that could never work for anyone in OAuth mode, since `Role#allowed_permissions` intersects the role with the token's scopes.

## What it deliberately does not do

`archive` / `unarchive` are `accept_api_auth`, but `ProjectsController` gates them on `require_admin` rather than a project permission, and `admin` is never advertised — so no token this server issues could reach them.

`delete` is reachable (`delete_project`; the `require_sudo_mode :destroy` filter returns early for API requests) and is left out on judgement. It destroys every issue, wiki page, document and file in the project and in each of its subprojects, and advertising `delete_project` would put "Delete the project" on every consent screen. Say the word if you would rather it be symmetric with `manage_redmine_version` and I will add it.

## Upgrade note for OAuth deployments

The six new permissions must be ticked on the Redmine OAuth Application **before** the upgraded server starts, or new authorizations fail with `invalid_scope` while everything already authorized keeps working. The CHANGELOG entry says so. Their labels in Redmine's `en.yml` are Create project, Create subprojects, Edit project, Close / reopen the project, Set project public or private, and Select project modules.

## Verification

Every Redmine claim is checked against the released tag `6.1.1`, not a branch:

- `app/controllers/projects_controller.rb:25-35` — the filters and the `accept_api_auth` list
- `lib/redmine/preparation.rb:38-43`, `:47` — the project permissions
- `app/models/project.rb:825-873` — the safe-attribute blocks, including the two conditional ones
- `app/views/projects/show.api.rsb` — the response shape `_project_to_dict` mirrors
- `lib/redmine/sudo_mode.rb:155-165` — `SudoRequestFilter#before` returning early for `api_request?`

python-redmine 2.5.0 supplies all four endpoints: `Project.query_create` / `query_update`, and `ProjectManager.__getattr__` for `close` / `reopen`.

## Tests

`tests/test_manage_redmine_project.py`, 30 tests: the four actions, the two refusals (`archive` and `delete` as invalid actions), read-only mode on every action, required-argument and path-injection validation on `project_id` and `identifier`, `update` sending only the fields the caller set while keeping falsy ones, the create-only `identifier` rule, the read-back, and the serializer — including that an absent key comes back `None` rather than a fabricated `False`, and that `description` is wrapped as insecure content.

`tests/test_tool_annotations.py` count assertions move 63 → 64 and `WRITE_DESTRUCTIVE` 16 → 17. The scope table's own anti-drift tests in `tests/test_scope_enforcement.py` pass unchanged.

Docs updated: `docs/tool-reference.md` (full entry plus the read-only-mode list), `README.md`, `docs/contributing.md` and `pages/index.html` for the 49 → 50 core tool count.


jztan (changes requested):
@mmahmed thanks for this. The issue research made it easy to review, and I like that `update` refuses `identifier` instead of letting Redmine drop it silently. Two changes before I merge:

1. Read the project back with `include=["enabled_modules", "trackers", "issue_custom_fields"]` and add those three to `_project_to_dict`, on create as well as update/close/reopen. Right now only `is_public` shows whether a gated field took, so a dropped `enabled_module_names` never shows up in the response.
2. A few tool counts were missed: the summary line at `README.md:645` still says 49 and "Project Management (9)", and `docs/contributing.md:669` and `docs/roadmap.md:8` (also "maximum 63") still say 49.

Leaving out archive and delete is right, and I'm fine with the six new scopes shipping by default alongside the CHANGELOG note.

mmahmed:
Both done, and the branch is rebased on `develop` now that #306 has landed. Four things worth your eye before you re-review.

**1. The includes.** `update`, `close`, `reopen` and `create` all go through one read-back helper that asks for `include=enabled_modules,trackers,issue_custom_fields`, and the serializer carries all three. `create` needs its own `GET`: the `POST` renders `show.api.rsb` but carries no `include=` params, so the created resource has none of the arrays.

I checked the permission question before adding them, since requesting an include that turns out to be gated would have meant a scope change. `ProjectsHelper#render_api_includes` at 6.1.1 guards each of the three on `include_in_api_response?` alone, with no permission check, so no new scope. `TOOL_SCOPES` and `WRITE_SCOPES` are unchanged from what you already approved.

Two shape decisions I made rather than guessed at, both easy to flip if you disagree:

- `enabled_modules` comes back as **names** (`["issue_tracking", "wiki"]`), not `{id, name}` — it matches the `enabled_module_names` parameter that writes it, and `get_project_modules`' existing shape. `trackers` and `issue_custom_fields` come back as `{id, name}`, because `tracker_ids` and `issue_custom_field_ids` write them by id. So each collection reads back in the currency you write it in.
- All three are read with `_included_list`, not `getattr`. All three names are in python-redmine's `Project._includes`, so the attribute fires a second request instead of reporting the key absent — the trap `_included_list`'s docstring describes. There is a test that sets the attributes and leaves the payload empty; it fails if anyone switches to `getattr` later. That is also why `_enabled_module_names` is not reused here, even though it is the obvious helper: it reads through the attribute.

Your underlying point was right, and it is now the thing a test asserts directly: `test_a_dropped_enabled_module_names_write_is_visible` writes two modules, has Redmine store one, and checks the response shows one.

**2. One thing your change surfaced that you did not ask for.** Adding a read-back to `create` introduces #146's failure mode: the project exists once the `POST` returns, so if the follow-up `GET` fails, reporting an error invites a retry and a duplicate project. It is not hypothetical — `Project#add_default_member` only runs when a default role is configured, so a non-admin creator on an instance without one can create a project and then be refused `projects#show`. The read-back on `create` is therefore outside the error handler and never fatal: on failure it falls back to the `POST` body (everything but the three arrays) and logs a warning. `test_a_failed_read_back_does_not_turn_a_create_into_an_error` covers it. Say if you would rather it just propagate.

**3. The counts.** All three you listed were real misses. They have moved again since you wrote the review, though — #306 added a tool, so the core count is now **51**, not 50, and the total is 64 with a maximum of 65.

Rather than hand-count again I enumerated `mcp.list_tools()` and subtracted the plugin-tagged and admin-gated ones. Fixed in `README.md` (the summary line, the section heading, the feature bullet, and *Project Management (9)* → *(10)*), `docs/contributing.md` (both the total and the tree comment at the bottom), `docs/roadmap.md`, and `pages/index.html`.

**4. A rebase hazard worth knowing about.** `tests/test_tool_annotations.py::test_table_size_and_kind_counts` conflicted, and git resolved the `len(own)` line silently and wrongly: #306 and this branch both changed it `63` → `64`, so the merge treated two independent additions as one change and kept `64`. The real value is `65`. The per-kind assertions did conflict and so were caught, but that one line would have merged clean and wrong. It might be worth having that assertion derive from the table rather than restate a literal, but I have left it alone here rather than widen the diff.

CI is green on 3.10 through 3.14. Locally the suite's failure set is identical to `develop`'s at the same commit, with the added tests accounted for (2810 passed against the baseline's 2771).

jztan:
@mmahmed merged, thanks! Checking `render_api_includes` for a permission gate before adding the arrays, and spotting the `len(own)` count that git merged wrong in the rebase, both saved me time. I'll handle the create read-back fallback in a follow-up.

PR #312, opened 2026-09-19
feat(extensions): let an extension widen the issue tools

### LINKED ISSUE #311: Extension point for issue update keys and query filters
### Problem Statement

#295 gave an out-of-tree module a door for *new* tools. A distribution or plugin that extends an *existing* resource still has none, and the issue tools are where that bites: two closed tables decide what `update_redmine_issue` may write and what `list_redmine_issues` may filter on.

**Writing.** `_STANDARD_ISSUE_UPDATE_FIELDS` in `_custom_fields.py` is a module constant, and a key that is not in it is treated as a custom field *label*. That is not merely a miss, it is a collision: `_normalize_field_label` reduces a label to `[a-z0-9]`, so a custom field named "Easy Sprint ID" normalizes to the same `easysprintid` as the attribute `easy_sprint_id`. An extension cannot work around this from outside either, because the workaround would have to be "do not name a custom field like the attribute", which is not the extension's to decide.

**Filtering.** `_ISSUE_QUERY_FILTER_NAMES` in `tools/issues.py` feeds `_reject_unregistered_filter_keys`, and that refusal is worth keeping: Redmine drops a filter it does not know and answers 200 with the collection unnarrowed, which a caller cannot tell apart from a filter that matched everything. A distribution that registers its own query filters therefore needs a way to widen the list, not a way to switch the check off. Some of those filters also need a companion parameter — Easy Redmine's EasyIssueQuery engages on `set_filter=1` — and there is no seam for that either.

Reading is deliberately not part of this request: `unmapped_fields` (2.15.0) already passes these keys through, nested objects included. That half is solved.

What is left is small and specific, which is the point of raising it: for our deployment the remaining reason to keep a fork of this repository is ~120 lines in two of your files. Everything else we carry is either upstream already (api-key login) or moves into a private extension package under #295 (six tools for Easy Redmine sprints, attendance and checklists).

### Proposed Solution

Two more `ExtensionSpec` fields, enforced the way #295 enforces the rest:

```python
    family="easy",
    enabled=lambda: is_true_env("REDMINE_EASY_ENABLED"),
    tool_kinds={...},
    tool_scopes={...},
    issue_update_keys=("easy_sprint_id", "easy_story_points", "target_backlog"),
    issue_query_filters={"easy_sprint_id": {"set_filter": 1}},
))
```

Semantics:

- Both apply only while that extension's `enabled()` is true. With no extension loaded, or every one disabled, both tables are exactly what they are today and the request this server sends is byte-identical to the current one.
- `issue_update_keys` join the standard-field set for the duration, so they reach the PUT as sent and never enter custom-field label matching.
- `issue_query_filters` keys join the accepted filter names. The mapped dict is merged into the API query with `setdefault`, and only when at least one filter of that extension is actually present in the call — a stock request never grows a parameter because an extension happens to be loaded.
- Collisions raise `RuntimeError` at import, like the existing checks: a key already in `_STANDARD_ISSUE_UPDATE_FIELDS`, a filter already in `_ISSUE_QUERY_FILTER_NAMES`, a filter name registered by another extension, or a companion parameter that `_reject_reserved_query_keys` reserves.
- Nothing bypasses validation. `_reject_non_scalar_filter_values` and the rest run unchanged; an extension declares *names*, never behaviour.
- The two inventory assertions get the same treatment #295 gave the annotation tables: subtract extension-registered entries so they keep measuring this tree.

A narrower shape works for us too, if you prefer it — for instance a plain set of filter names plus a separate `issue_query_params` mapping, or update keys only in a first step. The direction matters more to us than the spelling.

### Use Case

Easy Redmine is a fork of Redmine rather than a plugin: it serves the standard `/issues.json` with extra attributes and registers extra query filters. `easy_sprint_id`, `easy_story_points` and `target_backlog` are attributes of its own `IssueApiRequest`, so they ride the normal issue update and need no endpoint of their own; `easy_sprint_id` is an EasyIssueQuery filter, which is how "the issues in this sprint" is asked for.

With these two fields, our fork of this repository drops to zero lines in your files. Our users keep one connector, one consent and one set of issue tools — "move this ticket to the current sprint" stays `update_redmine_issue`, not a second tool that exists only because the attribute is ours.

The mechanism is general. Any distribution or plugin that adds attributes or filters to issues needs exactly this, and none of it is specific to Easy Redmine.

### Feature Category

Other

### Priority

Important

### Alternatives Considered

- **Parallel extension tools** (`update_easy_issue`, `list_easy_issues`). Works today with no change from you, and we have the code. Rejected because it gives a model two ways to change one issue and splits the workflow by who owns the attribute rather than by what the caller wants — plus a second copy of the validation, the read-only refusal and the scope mapping, which is the class of duplication that drifts.
- **Opening the filter list**, accepting unknown names and letting Redmine decide. Rejected: it throws away the protection that makes the closed list worth having, the silent 200 with an unnarrowed collection.
- **Keeping the fork.** ~120 lines in two files, rebased on every release, with the forked behaviour sitting outside the tests that make the tables mandatory.

### Additional Context

The reading half of the same problem is already solved by `unmapped_fields` (2.15.0), which is why this asks only for the write and filter tables.

We have the forked implementation running (a `REDMINE_EASY_ENABLED` flag, the three update keys, the one filter plus `set_filter=1`), so the PR would be that code turned inside out into the extension surface, with tests for: both tables untouched when nothing is registered, the companion parameter appearing only when one of the extension's filters is in the call, each collision raising at import, and the existing validation still rejecting what it rejects today.

### Contribution

andilem:
## Summary

Fixes #311.

`ExtensionSpec` gains two fields, so an extension can widen the issue tools instead of shipping parallel ones next to them: `issue_update_keys` for attributes `update_redmine_issue` may write, and `issue_query_filters` for filter names `list_redmine_issues` accepts, each mapped to the query parameters that have to accompany it.

Both are read from the registry per call and only for families whose `enabled()` is true. With no extension loaded, which is the default, both tables are exactly what they are today and the request this server sends is unchanged — pinned by a test that asserts the parameter set of a stock call.

Neither seam could be worked around from outside the server. An update key it does not know is taken for a custom field *label*, and labels are matched with every non-alphanumeric stripped, so the attribute `easy_sprint_id` and a custom field named "Easy Sprint ID" normalize to one string and the value is written to whichever the lookup found. A filter name it does not know is refused — rightly, because Redmine drops an unregistered filter and answers 200 with the collection unnarrowed, which a caller cannot tell apart from a filter that matched everything. So the filter list is widened, never opened.

## Changes

- `extensions.py`. Two `ExtensionSpec` fields with their shape checks (`_check_issue_update_keys`, `_check_issue_query_filters`; a bare `str` of update keys is refused for the same reason an advertised-scope list is), and `_check_issue_seams()` run by `register_extension` before anything merges.
- `_extension_registry.py`. `extension_issue_update_keys()` and `extension_issue_query_filters()`, both walking the enabled families only. The registry keeps importing nothing but `_env`.
- `_custom_fields._is_standard_issue_update_key()` consults the registry. That one predicate is also what decides whether the update path fetches the project's custom fields at all, so a registered attribute additionally spares the call a round-trip.
- `tools/issues.py`. `_reject_issue_filters` accepts the registered names in addition to `_ISSUE_QUERY_FILTER_NAMES`, and `list_redmine_issues` merges a filter's companion parameters with `setdefault`, only when that filter is present in the call.
- `docs/extensions.md`: a section on extending the issue tools, and the new refusals added to "What is checked at startup". `docs/tool-reference.md`: one bullet in the Extensions list. CHANGELOG entry under Unreleased.

Registration refuses what it cannot merge, at import, in the style of the existing checks: a key Redmine itself defines, a filter it itself registers, either one already claimed by another family, a companion parameter Redmine reads as the query's own filter definition (`fields`, `f`, `query_id`), and one `list_redmine_issues` owns (`limit`, `offset`, `sort`, `include`). Filter *values* keep facing the existing validation — an extension declares names, never behaviour.

## Testing

```
flake8 src/ --max-line-length=88       clean
black --check src/ --line-length=88    clean (51 files)
pytest -m "not integration" tests/     green
pytest -m integration tests/           skips without a Redmine; one
                                       pre-existing failure, see below
```

`tests/run_tests.py` itself shells out to a bare `python`, which on this Windows box resolves to the launcher rather than the venv, so I ran the two selections it makes directly.

`tests/test_extension_issue_seams.py`, new, 35 tests: every shape rejection; each collision (Redmine's own names, another family, reserved keys, the tool's own parameters); a rejected spec leaving the registry untouched; the flag read per call and two families merging; the predicate and the skipped custom-field lookup; the label collision the update seam exists for, asserted in both directions; an unregistered filter still refused, a registered one accepted, refused again once the family is disabled, and its values still validated; the companion parameter riding along only with its filter and never over the caller's value; a stock call's parameter set unchanged; and the whole path once through the MCP surface with `fastmcp.Client`.

Two failures in my environment are unrelated to this branch and reproduce on `develop` with the change stashed: `tests/test_extensions.py::TestColdStartImport` (its subprocess cannot initialize Winsock here, `OSError: [WinError 10106]`) and `test_ssl_integration.py::test_ssl_cert_file_exists`.

## Checklist

- [x] `python tests/run_tests.py --all` passes (activate `.venv` first) — as the two pytest selections it makes; see Testing
- [ ] Change works with both local (`uv run python -m redmine_mcp_server.main`) and Docker (`docker-compose up`) deployments — local only. The change is inert for a deployment that loads no extension, and nothing here touches packaging or startup, so I did not build an image for it.

## Notes for review

Three choices worth disagreeing with explicitly:

- **`setdefault` for the companion parameters, not an override.** An extension can register a parameter as a filter of its own, and then a caller can set it; this keeps the caller's value. The alternative would let a spec silently rewrite an argument someone passed.
- **Refusing a key Redmine already defines** rather than ignoring it. It is harmless to merge, but the case it catches is a release where the built-in support caught up with the extension, and a spec that says so is worth being told about.
- **Local imports in `_check_issue_seams`.** `tools.issues` reaches the whole tool tree, and `extensions` is imported from inside an extension module during `main`'s import loop. By then `tools` is imported, so the lookup is free; a module-level import would be a second path into that tree.


jztan (changes requested):
@andilem Thanks, this sits well next to #295, and widening the filter list instead of opening it was the right call. Two changes before I merge:

1. `_is_standard_issue_update_key` is shared with `create_redmine_issue`, so registered keys pass through on create as well. I want that behaviour. Say so in `docs/extensions.md` and the `issue_update_keys` docstring, and add a create test.
2. Refuse a companion parameter that is itself a filter name, built-in or registered by any family. Right now `{"acme_sprint_id": {"status_id": "open"}}` registers fine and quietly narrows every call that uses the filter.

andilem:
Both done in 8549ef4.

**1. Create-path pass-through.** You are right that it was already the behaviour and wrong that nothing said so. It is now in the `issue_update_keys` docstring, the "Extending the issue tools" section, the tool reference and the changelog, with two tests: a registered attribute reaches `issue.create` as an attribute *and* the project's custom fields are never fetched (there is nothing left to rule out), and the same call without the registration lands in the custom field whose label normalizes to the attribute's name.

**2. Companion parameter that is a filter name.** Refused at registration now — Redmine's own names, the spec's own filters, and any other family's. Your `{"acme_sprint_id": {"status_id": "open"}}` raises instead of quietly dropping every closed issue from a sprint listing.

One consequence worth flagging: that rule removes the only case the `setdefault` merge protected against, since a companion parameter can no longer collide with anything a caller is able to set. I kept `setdefault` anyway — it is what makes the collision a registration-time concern rather than a runtime one — but the test that exercised a caller overriding a companion is gone, because the shape it needed is now rejected.

Suite green, `flake8 src/` and `black --check src/` clean.

jztan:
@andilem Thanks for the quick turnaround, both changes are what I wanted, and your `setdefault` reasoning holds. Two holes left in the companion check:

1. A `cf_<id>` companion still registers: `{"acme_sprint_id": {"cf_3": "x"}}` goes through, and `cf_3` is a filter `list_redmine_issues` accepts. Match companions against `_CUSTOM_FIELD_FILTER_PATTERN` and the association custom-field pattern as well as the name sets.
2. The check only runs one way. Family A registers a companion `foo`, then family B registers `foo` as a filter, and both pass. Refuse a new filter name that an already registered family uses as a companion.

Add a test for each.

andilem:
Both closed in e7724a8.

**1. `cf_<id>` companions.** Right, and the reason it slipped is exactly what you name: those filters are registered per custom field, so there is no name set to intersect with. Companions now go through `_is_custom_field_filter` with the issue query's own associations, which covers `cf_3`, `cf_3.due_date` and `author.cf_3` in one place rather than a second copy of the patterns. Parametrised test over all three spellings.

**2. The reverse direction.** Also right. A new spec's filter names are checked against every companion the registered families already send, and the message names the family that holds it, since that is the one you would have to look at. Test registers A with companion `acme_mode`, then B with `acme_mode` as a filter.

Suite green, `flake8 src/` and `black --check src/` clean.

jztan:
@andilem Merged, thank you. Reusing `_is_custom_field_filter` for the `cf_<id>` check instead of copying the patterns was a nice touch. Hope this lets you drop the fork.

PR #315, opened 2026-09-19
feat(journals): report a long field-change value by length, not in full

### LINKED ISSUE #313: Journal details echo every past description in full: 94% of an 879k-character issue response, paid on every read
`get_redmine_issue(include_journals=True)` on a real ticket returned **879,251 characters** — enough that the client refused the result outright and spilled it to a file. Measured composition:

| part | chars | share |
| --- | ---: | ---: |
| journal `details` | 827,333 | **94.1 %** |
| — of which `attr/description` | 827,044 | 14 entries |
| the description itself | 42,561 | 4.8 % |
| **all comment text** | **4,704** | **0.5 %** |
| everything else (subject, status, custom fields, …) | 1,411 | 0.2 % |

18 journals, 16 of them field changes with no note. The largest single detail carries `old_value` of 26,917 characters and `new_value` of 51,857.

The cause is Redmine's own data model: a description change is journalled with the full old *and* new text in `journal_details`, and `include=journals` hands all of them back. Fourteen edits of a ~50 KB description therefore produce ~827 KB of change log, against 4.7 KB of what anyone actually wrote. The web UI never shows this — it renders a diff link — so it is invisible until something reads the API.

## Why this is worth fixing here

**Cost.** That response is roughly 220–250k input tokens (at ~3.5–4 chars/token). Not once: every read of the ticket. And the client's oversize handling makes it worse rather than better — it spills to a file and then instructs the model to read the whole file back in sequential chunks for any summarization, analysis or review. So the tokens are paid anyway, plus the round trips.

**Latency.** Fetching, serialising and re-reading ~880 KB dominates the call. A question like "what is the current status" costs the same as a full review.

**It compounds.** Every further edit of the description adds another ~100 KB to every future read. A ticket that is actively maintained gets monotonically more expensive to look at, and nothing ever reclaims it.

**None of it is information a caller usually wants.** The current values live on the issue itself, not in the details; what a reader needs from a detail-only journal is *that* the description changed, when, and by whom.

## Proposal

1. **Elide long `old_value` / `new_value` in journal details by default.** Above a threshold (a few hundred characters), return the lengths and a flag instead of the text:

   ```json
   {"property": "attr", "name": "description",
    "old_value_length": 26917, "new_value_length": 51857, "elided": true}
   ```

   An explicit opt-in (`include_journal_values=True`, or a per-journal fetch) can still return the full text for the rare caller that wants it. For the ticket above this is 879,251 → ~52,000 characters, a 94 % cut with nothing lost that a caller normally reads. There is a single choke point: `_journal_details_to_list` in `tools/issues.py`, which both `_journals_to_list` and `_journal_to_dict` go through.

2. **`journal_order="desc"`.** Journal pagination exists but runs oldest-first, so `journal_limit=5` returns the *oldest* five — the opposite of what a caller asking for recent history wants, and the reason our own house rules tell people not to use it. Newest-first would make the existing parameter usable, and "the last five comments" is the common case.

3. **Field selection on `get_redmine_issue`.** `list_redmine_issues` already takes `fields`; the single-issue tool has only `include_*` switches, so there is no way to ask for subject and status without the description.

(1) is the one that matters; (2) and (3) are small and independent.

Happy to open the PR.

---

Sibling: #314 covers the other direction — writing a long description forces the model to retype it. The two are independent (Redmine journals the result, not the patch), and the same oversized descriptions drive both.

andilem:
Closes #313. Implements proposal (1) only — the eliding, which is the part that pays.

The other two proposals from that issue now have their own issues, so closing #313 with this PR loses nothing:

- #318 — `journal_limit` returns the *oldest* journals, with the measurement
- #319 — no field selection on `get_redmine_issue`, so a large description is paid on every read

## The problem, measured

`get_redmine_issue(include_journals=True)` on a real ticket returned **879,251 characters**:

| part | chars | share |
| --- | ---: | ---: |
| journal `details` | 827,333 | **94.1 %** |
| — of which `attr/description` | 827,044 | 14 entries |
| the description itself | 42,561 | 4.8 % |
| **all comment text** | **4,704** | **0.5 %** |

Redmine journals a description edit with the full old *and* new text, so fourteen edits of a ~50 KB description produce ~827 KB of change log. Roughly a quarter of a million input tokens, on **every** read of the ticket, and it grows with each further edit. Invisible in the web UI, which renders a diff link instead.

## The change

A value longer than `REDMINE_MCP_JOURNAL_VALUE_MAX_CHARS` (default 500) is reported by length:

```json
{"property": "attr", "name": "description",
 "old_value": null, "new_value": null,
 "old_value_length": 26917, "new_value_length": 51857, "elided": true}
```

- **The value keys stay present** and hold `null`, so a caller reading them gets no `KeyError`, and `elided` tells an omitted value apart from an empty one.
- **Each value is judged on its own** — a short `old_value` survives beside a long `new_value`.
- **Lengths are measured before `wrap_insecure_content`**, so they report the field's own size rather than the boundary tags'.
- **Opt back in per call:** `get_redmine_issue(..., include_journal_values=True)` and the same on `get_private_notes`, for when the *previous* text is what is being asked for. `REDMINE_MCP_JOURNAL_VALUE_MAX_CHARS=0` disables eliding entirely.

Nothing a caller normally reads is lost: the current values are on the issue, and what a reader wants from a field change is that it happened, when, and by whom.

One choke point does the work — `_journal_details_to_list`, which `_journals_to_list` and `_journal_to_dict` already both went through.

## Tests

`tests/test_journal_value_eliding.py`, 12 cases: long values elided, short ones untouched, each value judged separately, `include_values` returning the text, threshold `0` disabling it, non-string values left alone, the env var's default and its fallbacks for empty/garbage/negative input, and a reconstruction of the ticket's shape asserting the response drops by more than 50× while the one real comment survives.

## Note on the default

500 characters is a judgement call, not a measurement. It is well above any enum, id, date or status value and well below a description worth reading. Say the word if you would rather have it higher, lower, or opt-in rather than opt-out — the knob and the per-call override are both already there.


jztan (changes requested):
@andilem thanks for measuring this before proposing anything, the table made the case on its own. I'm keeping eliding on by default. Two things before I merge:

1. `Closes #313` will close the issue while `journal_order="desc"` and field selection on `get_redmine_issue` are still undone. Please file each as its own issue so they don't get lost.
2. Please add `REDMINE_MCP_JOURNAL_VALUE_MAX_CHARS` to the environment variable table in the README.

andilem:
Both done.

**1.** The remaining two proposals from #313 are now their own issues, so `Closes #313` loses nothing:

- **#318** — `journal_limit` returns the oldest journals. Carries the measurement: journal ids and dates from the live ticket, and the point that the mechanism is sound and cheap (88,911 characters instead of 879,251) but aimed the wrong way. "The last five comments" needs `total`, and `total` needs the full response.
- **#319** — no field selection on `get_redmine_issue`. This is the one this PR leaves behind: after eliding, that ticket is down to ~54,900 characters and **42,561 of them are the description itself, 78% of what remains**. A caller asking for the status pays it in full.

The PR description now links both.

**2.** `REDMINE_MCP_JOURNAL_VALUE_MAX_CHARS` is in the README table in baa2351, next to the other attachment and cleanup knobs. The row names the default and both ways back — `0` to disable eliding entirely, `include_journal_values=true` for a single call.

jztan:
@andilem merged, thanks. The measurement in #313 is what made this an easy yes, and splitting out #318 and #319 keeps the rest of it on the list.

Issue #314, opened 2026-09-19
Editing a long description makes the model retype it: slow, costly, and text goes missing


andilem:
Redmine replaces `description` wholesale — there is no patch endpoint — so `update_redmine_issue` takes the finished text. For a description of 50–100k characters that means the model has to **write all of it out**, character by character, into a tool argument. We have tickets in that range, and editing them is slow, expensive, and occasionally lossy.

This is the same failure as #305, one order of magnitude up. There the payload was 4,312 characters of base64 and 32 of them went missing in a repetitive stretch, silently attaching a corrupt PNG. At 50k characters of prose the same transcription is being asked for, and text does go missing.

## The three costs

**Output tokens, and therefore latency.** ~50k characters is roughly 14k output tokens. Output is generated sequentially and is the slow part of any model call — this is what makes an edit take minutes rather than seconds. It is also the expensive direction.

**Input tokens.** The current description has to be read before it can be changed, and it is read again with every subsequent look at the ticket. (The much larger half of that is #313: past descriptions echoed in `journal_details`.)

**Correctness.** Nothing verifies that what arrived is what was meant. A dropped paragraph in the middle of a 50k description looks exactly like an intentional edit, and it goes out to watchers by mail immediately.

## Proposal

**1. Patch instead of replace — the change that matters.**

Let `update_redmine_issue` take edits rather than the finished text:

```json
{"description_edits": [{"find": "…the passage to replace…", "replace": "…the new text…"}],
 "description_expected_sha256": "…"}
```

Applied server-side, each `find` required to match exactly once, the whole call refused otherwise. The model writes a few hundred characters instead of 50,000: output shrinks by two orders of magnitude, latency with it, and there is nothing long to miscopy. This is what the Edit tool does for files, and for the same reasons.

`description_expected_sha256` matters more here than the checksum in #305 did: between reading a description and patching it, someone else may have edited it. A guard turns a lost-update into a refusal. It is also cheap — the server already has the current text in hand.

**2. `description_upload_id`, for when the text really is all new.**

`create_upload_ticket` and `POST /uploads/{id}` landed in #306. Accepting a staged `upload_id` as the source for `description` would let a whole replacement travel the same way an attachment does — from disk to the server, never through the model. This composes neatly with how large results already come back: an oversized response is written to a file the model can edit in place, and this is the missing way back.

**3. The same for `manage_issue_note(action="edit")`.**

[...]

PR #316, opened 2026-09-19
feat(issues): set a long description without writing it out — patch it, or send it as a file

andilem:
Addresses #314, parts 1 and 2 — patching, and taking a whole new description from a staged upload. Part 3 is left out; see **Scope** below.

## The problem

Redmine replaces `description` wholesale, so `update_redmine_issue` takes the finished text — and that text is written out by the model, character by character, into a tool argument. For a 50–100k-character description that means:

- **Output tokens, and therefore latency.** ~14k output tokens for a 50k-character field, generated sequentially. This is what makes an edit take minutes.
- **Correctness.** It is the same transcription as #305, where 32 characters went missing out of 4,312 and silently attached a corrupt PNG. At 50k characters of prose, text does go missing — and a dropped paragraph in the middle of a description looks exactly like an intentional edit, mailed to watchers immediately.

## The change, part 1: patching

```python
update_redmine_issue(
    issue_id=123,
    fields={},
    description_edits=[{"find": "runs on two nodes", "replace": "runs on four nodes"}],
    description_expected_sha256="<digest of the description as it was read>",
)
```

The model writes the passage that changed instead of the document. Output drops by orders of magnitude, latency with it, and there is nothing long left to miscopy.

**Each `find` must occur exactly once** in the text as it stands after the preceding edits. Zero matches means the caller is working from a version that is not the current one; more than one means the edit is ambiguous and could land in the wrong place. Both refuse the whole call and leave the issue untouched — a half-applied patch is worse than none — and the error names the index of the pair that failed. `description_edits` and a `description` in `fields` together are refused rather than one silently winning.

**`description_expected_sha256`** guards the gap between reading a description and patching it. It matters more here than the checksum in #305 did: there a mismatch meant a corrupt file, here it would mean silently discarding someone else's work. Optional, checked before any edit, and cheap — the server has the current text in hand either way.

## The change, part 2: a staged upload as the source

#306 has merged, so this is no longer blocked. Where the description really is all new rather than edited, `description_upload_id` takes it from a file staged with `create_upload_ticket`, decoded as UTF-8 — the content travels from disk to the server the way an attachment does. Non-UTF-8 bytes are reported rather than mangled.

This composes with how an oversized read already comes back: a client that spills a large result to a file gives the caller something to edit in place, and this is the missing way back.

`description`, `description_edits` and `description_upload_id` are mutually exclusive, and passing more than one is refused with the names of those given rather than one quietly winning.

## Scope

- **`manage_issue_note(action="edit")`** is not included. Redmine serves no endpoint for a single journal, so *patching* a note needs the issue id and a full issue fetch — a design question of its own rather than a line of code. `_apply_text_edits` already takes the field name as a label so its errors read `notes_edits[1]` when that lands, and there is a test pinning that. Worth noting the two halves split unevenly: a `notes_upload_id` would need none of that, since setting a note from a file does not require reading the current one. Say the word and I will add it.

Happy to follow up, in whichever order suits you.

## Tests

`tests/test_description_patching.py`, 27 cases: a single edit landing, edits building on each other in order, deletion via an empty `replace`, the payload being orders of magnitude smaller than the document, absent and ambiguous `find` refused, a later failing edit discarding the earlier ones, unusable input shapes, the digest matching, mismatching, being optional, tolerating case and whitespace, and rejecting a non-string — plus four through `update_redmine_issue` itself asserting that only the finished text reaches Redmine, that a refusal never calls `issue.update`, and that edits plus a full description are rejected. Five more cover the staged-upload source: the text arriving intact including non-ASCII, non-UTF-8 bytes reported, an unknown upload id reported, two sources at once refused without touching Redmine, and the staged text reaching `issue.update` as the description.

## Note

This does not help #313 and #313 does not help this: Redmine journals the *result* of an edit, not the patch, so a patched description still writes its full old and new text into `journal_details`. Both are needed.


andilem:
Scope grew by one commit now that #306 has merged: `description_upload_id` takes a whole new description from a staged upload, which the original description of this PR listed as blocked. Patching still covers editing; this covers replacement. The title and body are updated accordingly.

Part 3 of #314 remains out, but the two halves of it split unevenly and the body now says so: a `notes_upload_id` needs nothing new, while `notes_edits` needs the issue id and a full fetch because Redmine serves no endpoint for a single journal.

_🤖 Added by [Claude Code](https://claude.com/claude-code)_

jztan (changes requested):
@andilem thanks, patching is the right shape for this, and refusing the whole call on a bad `find` is what I want. Two things don't work yet from an MCP client:

1. A caller can't compute `description_expected_sha256`. `get_redmine_issue` returns the description inside `<insecure-content-...>` tags with a random boundary, so a digest of what was read never matches the raw text. Please return `description_sha256` from `get_redmine_issue` (and in the mismatch error) so the caller only echoes it back, and add a test that goes through the read tool instead of hashing the raw string.
2. A multi-line `find` written with `\n` fails against a description stored with `\r\n`, which is what the web UI saves. Please match with line endings normalized, and keep the stored line endings on write.

Please also open part 3 of #314 (note editing) as its own issue, so #314 can close with this PR.

andilem:
Both fixed in b0d704e, and part 3 is now #317. Thanks — 1 in particular meant the guard refused every patch that tried to use it, so the feature only worked with the safety off.

**1. The digest was uncomputable.** Confirmed: `wrap_insecure_content` regenerates the boundary id on every call, so even a caller hashing the exact bytes it received would miss twice over. `get_redmine_issue` now returns `description_sha256`, taken from the raw description, and the caller echoes it back. The mismatch error names the current digest too, so a retry does not need a second read.

One judgement call worth flagging: I put it on the tool, not in `_issue_to_dict`. That serializer also backs `list_redmine_issues` and the search tools, where a digest per row is noise and changes a documented key set — sixteen shape tests made the point when I tried it there first. The cost is that `create_redmine_issue` and `update_redmine_issue` do not return it, so a caller patching twice in a row re-reads. Say the word if you would rather have it on those two as well.

The test goes through the read tool as asked, and asserts the thing that was broken — `result["description_sha256"] != sha256(result["description"])` — then feeds the returned digest into `update_redmine_issue` end to end.

**2. Line endings.** Also confirmed against a real ticket: Redmine's UI saves CRLF, and our own descriptions are stored that way. Matching now normalizes both sides, and the stored convention is restored on write, so patching one passage does not rewrite every line ending in the field. Covered from both directions: an LF `find` matching CRLF text, CRLF surviving the edit *including inside caller-supplied `replace` text*, an LF document staying LF, and the digest guard still seeing the raw stored text rather than the normalized one.

**3. #317** covers note editing. It splits unevenly and the issue says so: `notes_upload_id` needs nothing new, while `notes_edits` needs the note's current text and Redmine serves no endpoint for a single journal — so it needs an `issue_id` parameter and a full issue fetch. I laid out three shapes and recommended requiring `issue_id`, but that is a call for you rather than something to decide in a commit.

jztan:
@andilem merged, thanks for the quick turnaround. Keeping `description_sha256` on `get_redmine_issue` only was the right call, and restoring CRLF inside the `replace` text is a detail I would have missed asking for. I'll pick up the `issue_id` question in #317.

Issue #319, opened 2026-09-20
get_redmine_issue has no field selection, so a large description is paid on every read


andilem:
Split out of #313 so that issue can close with #315, which covers the eliding.

`list_redmine_issues` and the search tools take `fields` to narrow what comes back. `get_redmine_issue` has only `include_*` switches, so there is no way to ask for subject and status without also receiving the description.

That matters once a description gets large. On the ticket #313 measured, eliding the journal details brought the response from 879,251 characters to about 54,900 — and **42,561 of what remains is the description itself, 78% of the response**. A caller that wants the current status pays it in full, on every call, and so does one that only wants the last comment.

## Proposal

`fields` on `get_redmine_issue`, with the semantics `list_redmine_issues` already has, so the two read tools narrow the same way:

```python
get_redmine_issue(43376, fields=["id", "subject", "status", "assigned_to"])
```

`_issue_to_dict_selective` already exists and is what `list_redmine_issues` routes through for exactly this; the single-issue tool would use the same helper. `None` and `["*"]` keep today's behaviour, so nothing changes for existing callers.

Two details worth deciding rather than assuming:

- **`description` in the field list.** It is the expensive one, so the natural thing is that omitting it omits the text. Whether `description_sha256` (added in #316 for patching) should still come back when `description` is excluded is a real question — I would say yes, since it is 64 characters and it is what makes a patch possible without reading the text at all.
- **`journals` and `fields` together.** They are separate switches today. I would leave them separate rather than folding journals into `fields`, since `include_journals` already carries pagination parameters of its own.

Happy to open the PR once the shape is settled.

jztan:
@andilem agreed on `fields` with the `list_redmine_issues` semantics, reusing `_issue_to_dict_selective`. Go ahead once #318 is in, since both touch `get_redmine_issue`.

1. Always return `description_sha256`, even when `description` is left out. As you say, it's what lets a caller patch without reading the text.
2. Keep journals out of `fields`. The `include_*` switches stay independent, the same way `include_custom_fields` works in the list tool. One catch: `include_journals` and `include_attachments` default to `True`, so `fields=["id", "status"]` on its own still returns every journal and attachment. Say in the docstring that a cheap read needs both set to `False`.
3. Reuse the `fields` guard from `list_redmine_issues` so both tools reject the same bad input.

PR #320, opened 2026-09-20
feat(issues): journal_order, and sort journals before slicing them

### LINKED ISSUE #318: journal_limit returns the oldest journals, so "the last five comments" needs the full response first
Split out of #313 so that issue can close with #315, which covers the eliding.

`get_redmine_issue` paginates journals oldest-first, which is the opposite of what a caller asking for recent history wants. Measured on a ticket with 18 journals, ids and dates both ascending in the full response:

```
get_redmine_issue(43376, journal_limit=3)

journal_pagination: {"total": 18, "offset": 0, "limit": 3, "count": 3, "has_more": true}

returned:      514197 (2026-09-17T08:59:09)
               514198 (2026-09-17T09:00:05)
               514277 (2026-09-17T10:47:06)

newest three:  514928 (2026-09-19T16:16:22)
               514936 (2026-09-19T20:12:07)
               514937 (2026-09-19T20:15:58)
```

In the code it is a plain slice over the order Redmine returns — `all_journals[offset : offset + journal_limit]` — and Redmine returns journals in creation order, so `[0:3]` takes the front.

**The mechanism is sound and cheap**: that call came back in 88,911 characters instead of 879,251. What it cannot do is answer "the last five comments", which is the question anyone actually asks. Today that needs `journal_offset = total - 5`, and `total` is only known from the full response — the one the caller was trying to avoid fetching. In our own house rules the parameter is simply forbidden for that reason.

## Proposal

`journal_order: Literal["asc", "desc"] = "asc"` on `get_redmine_issue`. With `"desc"` the slice runs from the other end, newest first.

Smaller than it looks: `total`, `offset`, `count` and `has_more` are already computed and keep their current meaning, and the default preserves today's behaviour exactly. The only question worth settling is whether the returned list stays newest-first or is reversed back to chronological after slicing — I would return it as asked, since a caller that says `desc` is reading backwards anyway.

Happy to open the PR.

andilem:
Closes #318. Implements all three points, and the first one turned out to matter more than the issue did.

## The order was not ours to assume

You were right, and it makes the original report incomplete. `IssuesController#show` reverses an issue's journals for an API user with "display comments in reverse chronological order" set, and the code sliced whatever arrived. So `journal_limit=5` returned the five *oldest* or the five *newest* depending on the account behind the key — and the behaviour the docs described held only for accounts with the default setting. My measurement in #313 happened to be on such an account.

Journals are now sorted by id, which Redmine assigns in creation order, before anything else happens. `asc` and `desc` then mean the same thing for every caller.

This part is a **Fixed** entry in the CHANGELOG, separate from the feature.

## `journal_order`

`"asc"` (default, unchanged) or `"desc"`. With `"desc"` the newest come first and `journal_offset` counts from that end, so "the last five comments" is one call:

```python
get_redmine_issue(43376, journal_limit=5, journal_order="desc")
```

With `"asc"` that needs `journal_offset = total - 5`, and `total` is only known from the full response — the request the caller was trying to avoid.

Sorting and ordering apply with or without a limit. `journal_pagination` gains `order`, so a caller can see which end its offset counted from; the other counters keep their meaning. The default preserves today's behaviour for anyone not passing the parameter, beyond no longer depending on the account.

## Tests

`tests/test_journal_order.py`, 10 cases. The one that carries the fix: a reversed journal list — what an account with that display setting receives — comes back sorted, and a second asserts both orders agree whatever arrangement Redmine sent. Then `desc` returning newest first, `asc` unchanged, offset counting from the selected end, ordering without a limit, `order` echoed for both values, and the remaining counters unaffected.

#319 is next and rebases on this, as you asked.


jztan:
@andilem merged, thanks. The sort by id was the part I cared about, and your reversed-list test covers it. I'm folding your CHANGELOG entries into the existing sections on develop (the PR added a second `### Added` above the first), so you'll hit a small conflict there when you rebase #321.

Issue #317, opened 2026-09-20
Editing a long note has the same problem as a description, but patching it needs a journal read Redmine does not offer


andilem:
Split out of #314 so that issue can close with #316, which covers the description.

`manage_issue_note(action="edit")` replaces a note's text wholesale, exactly as `update_redmine_issue` did for the description before #316. A new comment is usually short and is genuinely the model's own output, so it is not the problem. *Editing* a long one is: the whole text has to be written out into the tool argument, which is slow and drops text at length — the failure #305 measured at 4,312 characters and #314 hit again at 50k.

The two halves of the fix split unevenly, which is why this is its own issue rather than a follow-up commit.

## The easy half: `notes_upload_id`

Setting a note from a file needs nothing that does not already exist. `create_upload_ticket` and `POST /uploads/{id}` landed in #306, `_text_from_staged_upload` in #316, and no current text has to be read. This is a parameter and three lines:

```python
manage_issue_note(action="edit", journal_id=514482, notes_upload_id="...")
```

## The awkward half: `notes_edits`

Patching needs the note's current text, and **Redmine serves no endpoint for a single journal**. `GET /issues/:id?include=journals` is the only way to read one, so a patch would need:

- an `issue_id` parameter on `manage_issue_note`, which today identifies a note by `journal_id` alone — the caller would have to supply both, and they can disagree;
- a full issue fetch per patch, which is the cost #313 is about. #313's eliding helps but does not remove it: the *current* description still comes along, and on the ticket measured there that is 42,561 characters to patch one comment.

`_apply_text_edits` already takes the field name as a label, so its errors read `notes_edits[1]` unchanged, and there is a test pinning that. What it needs is the read, and the read is the design question.

Three shapes, none obviously right:

1. **Require `issue_id`** and fetch with `include_journals=True, include_journal_values=False`. Simple, honest about the cost, and the caller usually has the issue id anyway since that is how it found the journal.
2. **Fetch by journal id alone** — not possible without an issue id; Redmine has no route.
3. **Have the caller pass the current text** as `notes_expected` and diff server-side. Avoids the read but puts the whole note back in the argument, which is the thing being avoided.

(1) looks right to me, with `notes_expected_sha256` as the same concurrency guard `description_expected_sha256` provides. Happy to implement it that way, or to do only `notes_upload_id` if the patching side is not worth the issue fetch.

jztan:
[...]

PR #321, opened 2026-09-20
feat(issues): field selection on get_redmine_issue

### LINKED ISSUE #319: get_redmine_issue has no field selection, so a large description is paid on every read

andilem:
Closes #319. **Stacked on #320** (#318) — both touch `get_redmine_issue`, and you asked for that one first. The diff here is just the last commit; merge #320 and this rebases to nothing.

## Why

#315 moved the cost rather than removing it. On the ticket #313 measured, the response went from 879,251 characters to about 54,900 — and **42,561 of what remains is the current description, 78%**. Every caller that only wanted the status pays it.

## The change

`fields` narrows the issue's own keys, with the meaning `list_redmine_issues` already gives it and through the same `_issue_to_dict_selective`, so the two read tools behave identically: `None`, `["*"]` and `["all"]` return everything, an unknown name is skipped rather than refused, and naming `custom_fields` or `relations` implies the matching flag.

```python
get_redmine_issue(
    43376,
    fields=["id", "subject", "status"],
    include_journals=False,
    include_attachments=False,
    include_custom_fields=False,
)
```

**Your point 2, made loud.** `fields` does not reach journals or attachments — and while writing the tests I hit a third: `include_custom_fields` also defaults to `True`. So a cheap read names **four** things, not three. The docstring, the tool reference and a test all say so, because `fields=["id", "status"]` on its own returning every journal is exactly the surprise you were guarding against.

**Your point 1.** `description_sha256` survives any `fields` value. A test asserts it is identical between the narrowed and the full read, so a caller can patch a description it chose not to fetch — otherwise #316 would be undone by narrowing.

**Your point 3.** Same helper, so the same input rules, rather than a second guard that could drift.

## Tests

`tests/test_get_issue_field_selection.py`, 10 cases. The one that states the case measures it: the same read with and without `fields`, differing by more than fifty times in size, digest identical across both. Then the wildcards, an unknown name skipped, journals and attachments ignoring `fields`, the four-switch cheap read, and the implied flags in both directions.

Full suite passes apart from eight `TestColdStartImport` cases that fail the same way on a clean `develop` here — a local asyncio import in the subprocess, unrelated.

## One loose end from #317

Point 4 of that review — `_apply_text_edits` hardcoding `description_sha256` in its mismatch message — was fixed after you merged #316, so it did not make it in. I have it as a commit and will bring it with the `notes_upload_id` PR, unless you would rather have it on its own.


jztan (changes requested):
@andilem thanks. Spotting that `include_custom_fields` also defaults to `True` was a good catch, and the digest surviving `fields` works the way I wanted. One bug blocks the merge:

1. `fields=["relations"]` returns `"relations": []` on an issue that has relations. The serializer is told `include_relations=True`, but the `includes` list for the Redmine request is still built from the `include_relations` argument alone, so `include=relations` is never sent. `list_redmine_issues` solves this with `want_relations`, which feeds both the request and the serializer. Please do the same here, and add a test asserting that the `include` passed to `issue.get` contains `relations` when only `fields` names it.
2. When you rebase on develop, put the CHANGELOG entry inside the existing `### Added` section. I merged the duplicate headers from #320 there, so the conflict will be in that spot.

andilem:
Both addressed in 36143ba.

**1. `fields=["relations"]` returned an empty list.** You had it exactly right: the serializer was told `include_relations=True` but `include=relations` never reached `issue.get`, and `_included_list` reports a key that was never included as empty rather than fetching it. `want_relations` now decides both and is computed above the include list, the way `list_redmine_issues` does it. `want_custom_fields` moved with it for symmetry — that one was never wrong, since custom field values come back unconditionally.

Four tests in `TestRelationsNeedTheRequest`, against a mock that carries relations only when the request asked for them. That detail is the point: a mock that always returns them cannot see this bug. Two of the four fail on the old code — the one you asked for (the `include` passed to `issue.get`), and one asserting the relations actually arrive. A third checks the flag alone still works, a fourth that naming neither leaves `relations` off the request, since the include is not free.

**2. CHANGELOG.** Merged `develop` (4440d43) and put the entry at the end of the single `### Added` section. The `journal_order` and journal-sort entries on this branch were duplicates of what #320 already landed, so they are gone rather than merged.

jztan:
@andilem merged, thanks. Making the mock return relations only when the request asks for them was the right way to test it.

PR #322, opened 2026-09-20
feat(issues): replace a long note from a staged upload

andilem:
First of the two PRs for #317, as agreed — the half that carries no design question. `notes_edits` follows.

## `notes_upload_id`

Editing a note had the same problem a description had before #316: the new text is written out by the model, character by character, into a tool argument. `notes_upload_id` takes it from a file staged with `create_upload_ticket` instead, decoded as UTF-8. It needs nothing new, because setting a note does not require reading the current one — that is what separates this from the patching half.

```python
manage_issue_note(action="edit", journal_id=514482, notes_upload_id="...")
```

**Your point 3.** The response reports `notes_length` and `notes_sha256` rather than echoing the note, since echoing it would put the whole text in the conversation after all. A test asserts the response is an order of magnitude smaller than the note it wrote.

A plain `notes` call is unchanged and still echoes back — a caller that wrote the text already has it, and changing that would be a behaviour change for no gain. Tell me if you would rather have it uniform.

`notes` and `notes_upload_id` are mutually exclusive and one is required. Passing neither now names both in the error instead of only `notes`.

## Also here: point 4 of your review

`_apply_text_edits` hardcoded `description_sha256` in its mismatch message and pointed at `get_redmine_issue` — wrong for a note, whose digest will come from its journal. It now says `{label}_sha256` and leaves out where it is read from, which is the part that differs. That fix was pushed to #316 after you merged it, so it did not make it in; it rides here as its own commit (c3d9ae4) and a test asserts the notes label comes through.

## Tests

`tests/test_notes_upload.py`, 9 cases: the staged text reaching Redmine intact including non-ASCII, the response not carrying the note, non-UTF-8 and an unsent upload both refused without touching Redmine, the private flag still riding along, both sources at once refused, neither refused, and the two behaviours that must *not* change — a plain `notes` echoing back, and an empty string still clearing a note.

Full suite passes apart from eight `TestColdStartImport` cases that fail the same way on a clean `develop` here.

## For the next PR

Points 1 and 2 — requiring `issue_id` with `notes_edits`, rejecting a `journal_id` that is not one of that issue's journals in wording that also fits a private note the caller cannot see, and `notes_sha256` on each journal — land with the patching side. Noted that `include_journal_values=False` only trims our serializer and Redmine still sends the details; the fetch cost is what it is.


jztan:
@andilem merged, thanks. I merged develop into your branch first to sort out the CHANGELOG conflict, so pull before you touch #323. On your question: keep the echo on a plain `notes` edit. The caller already has that text, and dropping it would break existing callers for nothing.

PR #323, opened 2026-09-20
feat(issues): patch a note in place

### LINKED ISSUE #317: Editing a long note has the same problem as a description, but patching it needs a journal read Redmine does not offer

andilem:
Closes #317. **Stacked on #322** — the diff here is the last commit; merge that one and this rebases to nothing.

## `notes_edits`

`{find, replace}` pairs applied on the server, so a long note is changed by naming the passage rather than by writing the whole thing out. Same shape #316 gave descriptions, through the same `_apply_text_edits` — which already took the field name as a label for exactly this.

```python
manage_issue_note(
    action="edit",
    journal_id=514482,
    issue_id=43376,
    notes_edits=[{"find": "Stand: offen", "replace": "Stand: erledigt"}],
    notes_expected_sha256="...",
)
```

## Your four points

**1. `issue_id` only with `notes_edits`.** Required on that path, ignored on the others. A `journal_id` that is not among that issue's visible journals is refused before anything is written.

On the wording: an invisible private note is simply *absent* from the response rather than marked as hidden, so from here the two cases cannot be told apart. The message therefore does not claim the journal does not exist — it says the journal is not among the ones visible to you, adds that both ids should belong together, and names the "View private notes" permission as a reason a real id can still be missing. A test asserts the phrase "does not exist" is *not* in it.

**2. `notes_sha256` on each journal**, hashed from the raw notes, in both `_journals_to_list` and `_journal_to_dict` so `get_redmine_issue` and `get_private_notes` both carry it. Without it the guard had nothing to echo back — the same trap #316 hit: what a caller reads is wrapped in boundary tags with a fresh random id per call. A test asserts the journal's `notes_sha256` is *not* the digest of the `notes` the same response returned.

**3. `notes_length` and `notes_sha256` instead of the note**, on this path as on the upload one.

**4.** The label fix rides in #322 as its own commit.

## Tests

`tests/test_notes_edits.py`, 11 cases: only the changed passage reaching Redmine with the rest intact, the note not echoed back, an ambiguous `find` writing nothing, the error naming `notes_edits[0]` rather than the description, `issue_id` required here and not elsewhere, a journal not on the issue refused, that refusal allowing for an invisible note, the digest from the read tool accepted end to end, a stale digest writing nothing, and the three sources being mutually exclusive.

Full suite passes apart from eight `TestColdStartImport` cases that fail the same way on a clean `develop` here.

## On the cost

Both halves of that are yours and both hold: the fetch goes Redmine → server and never reaches the model, and `include_journal_values=False` only trims our serializer, so Redmine still sends the details. Nothing in this PR changes either — recording it here so the next reader does not mistake the eliding for a saving on the wire.


jztan:
@andilem merged, thanks. The refusal wording for a journal the caller can't see is exactly what I was after, and putting `notes_sha256` on `get_private_notes` too was the right call. That finishes #317.

Issue #325, opened 2026-09-20
A wiki page can only be replaced whole, so its text is written out by the model every time


andilem:
The same problem #314 solved for descriptions and #317 for notes, in the place where it should bite hardest: a wiki page is *the* long document in Redmine, and `manage_redmine_wiki_page` replaces its `text` wholesale.

There is no patch endpoint, so `text` carries the finished page — every character of it written out by the model into a tool argument. That is slow, because output is generated one token at a time, and lossy at length: #305 measured 32 characters going missing out of 4,312. A wiki page that is worth editing rather than rewriting is usually well past that.

Unlike an issue description, a wiki page is edited *by design* — that is what a wiki is — so the case for patching is stronger here than anywhere else in the API.

## Proposal

Mirror what the issue tools now have, with one change in Redmine's favour:

1. **`text_edits`** — `{find, replace}` pairs applied server-side, each `find` required to match exactly once, the whole call refused otherwise. `_apply_text_edits` already exists and already takes the field name as a label, so its errors would read `text_edits[1]` unchanged.
2. **`text_upload_id`** — take the whole page from a file staged with `create_upload_ticket`, for when it really is all new. `_text_from_staged_upload` already exists.
3. **The guard is `version`, not a checksum.** Redmine gives wiki pages optimistic locking of their own and answers 409 on a mismatch. Descriptions and notes needed an invented `*_expected_sha256` because Redmine offers nothing; here it would be reinventing something better that is already there. That also ties this to #324, which is about `version` not being sent on the text path today.

`get` would need to report the version alongside the text for a caller to echo back — it already returns `version` in `_wiki_page_to_dict`, so possibly nothing to do.

## We are not offering a PR for this one

We do not use the Redmine wiki. Anything we wrote would be untested against a real page, and the interesting decisions here — whether `version` is required or optional with `text_edits`, and whether the extra read is worth avoiding — need someone who actually edits wiki pages to judge. Reporting it because the shape is now established in two other tools and this is the third, not because it has bitten us.

Related: #324 (the version gap on the text path today).

jztan:
@andilem agreed, and `version` is the right guard. I'll build this after #324 so `text_edits` and `text_upload_id` can use the same `expected_version`.

Issue #324, opened 2026-09-20
A wiki update that changes the text sends no version, so a concurrent edit is silently overwritten


andilem:
`manage_redmine_wiki_page(action="update")` sends Redmine's `version` only when the caller is *not* changing the text. Writing new text sends no version, so a concurrent edit is overwritten without anyone noticing.

In `tools/wiki.py`:

```python
if text is None:
    # Sending the version read above turns a concurrent edit into a
    # 409 instead of a silent revert. Unchanged text does not create
    # a new revision.
    update_kwargs["text"] = existing_text
    update_kwargs["version"] = existing_version
else:
    update_kwargs["text"] = text
```

The comment states the reason exactly, and the protection lands on the branch that needs it least: an attachment-only update re-reads the page and echoes its text back, so it is guarded against reverting someone. The branch where the caller supplies new text — the one that actually replaces a page — sends no `version` at all.

`existing_version` is only read inside the `text is None` path, so on the text path there is nothing to send even if one wanted to.

## Why it matters more here than in a browser

Redmine's own wiki form carries the version in a hidden field, so a human editing in two tabs gets the 409. A caller of this tool does not, and a wiki page is exactly the kind of long document two people work on at once. The failure is silent: the losing edit is simply gone from the current revision, recoverable only from the history by someone who noticed.

## Proposal

On the text path, read the page first and send its `version`, the same way the attachment path already does — or accept an optional `expected_version` from the caller and send that, which saves the extra read when the caller has just read the page anyway. Redmine answers 409 on a mismatch, which is a better guard than a checksum because it is the server's own and needs nothing invented.

The existing read on the `text is None` path shows the shape; the ordering comment above it (do the cheap GET before `_build_upload_descriptors`) applies just as much to a read added here.

## We are not offering a PR for this one

We do not use the Redmine wiki, so anything we wrote would be untested against a real page and we could not judge the trade-off between an extra read and a new parameter. Reporting it because the code makes the gap plain, not because it has bitten us.

Related: #325 covers the other wiki gap, long text that has to be written out in full.

jztan:
@andilem confirmed, the text path sends no `version`. I'll fix it myself with an optional `expected_version`. A read inside the tool wouldn't help a caller whose copy of the page is already stale. `rename` has the same gap, so I'll fix that too.

PR #328, opened 2026-09-20
feat(issues): open an issue with a staged description

### LINKED ISSUE #326: create_redmine_issue cannot take a staged description, so a long one must be written out at creation
#316 gave `update_redmine_issue` a `description_upload_id`, so a long description can be set from a file staged with `create_upload_ticket` instead of being written out into the tool argument. `create_redmine_issue` did not get the same treatment:

```python
async def create_redmine_issue(
    project_id: int,
    subject: str,
    description: str = "",
    ...
```

So a long description can be *changed* without passing through the model, but not *written in the first place*. That is the wrong way round: at creation the whole text is new by definition, which is exactly the case the upload route exists for. Patching has no meaning here and is not being asked for.

This is not hypothetical for us. The issue that prompted #313 carries a 42,561-character description, and it was created through this tool — every character of it written out one token at a time, with no check that what arrived was what was meant.

## Proposal

`description_upload_id` on `create_redmine_issue`, with the semantics it already has on the update path:

- Mutually exclusive with `description`, refused with the names of both when given together.
- Resolved through `_text_from_staged_upload`, which already exists and already reports non-UTF-8 rather than mangling it.
- No checksum parameter: there is no prior text to guard against, and the bytes never passed through the model, so there is nothing to verify.

Small, and mostly a matter of routing the value into `fields["description"]` before the create call the way the update path does.

We use this tool and would open the PR.

andilem:
Closes #326.

#316 gave `update_redmine_issue` a `description_upload_id`, so a long description can be *changed* without passing through the model — but not written in the first place. That is the wrong way round: at creation the whole text is new by definition, which is exactly the case the staged upload exists for. Patching has no meaning here, and no checksum is asked for, since there is no prior text to guard and the bytes never reached the conversation.

Not hypothetical. The issue that prompted #313 carries a 42,561-character description, and it was created through this tool — every character produced one token at a time, with nothing to check what arrived against what was meant. The same failure `content_base64` has for attachments (#305), an order of magnitude up.

## What it does

```python
# ticket = create_upload_ticket(filename="description.html")
# curl -sS -H "X-Upload-Ticket: $TICKET" --data-binary @description.html "$UPLOAD_URL"
create_redmine_issue(
    project_id=1,
    subject="Cleanup-Review",
    description_upload_id="6f1e...",
)
```

The semantics are the update path's, minus the part that does not apply:

- **Mutually exclusive with `description`**, refused naming both. Checked before the uploads are resolved — an argument error costs nothing to report, and that step does I/O.
- **Emptiness is not a clash.** `description` is a string defaulting to `""`, so an explicit empty one cannot be told from an omitted one. Refusing that combination would only reject a call that asks for exactly one thing.
- **Resolved through `_text_from_staged_upload`**, which already reports non-UTF-8 rather than mangling it.
- **Read into the parameter the create call closes over**, so the required-custom-field retry sends it too rather than falling back to the empty default.

## Tests

Ten, in `tests/test_create_issue_upload.py`: the text arrives whole at 46k characters, other fields still go along, both sources at once are refused without creating, an empty `description` is not a second source, and a non-UTF-8 file, an unknown id and a ticket that never received a file each refuse the call rather than creating an issue with a wrong description. Two more assert the ordinary call is untouched.

Full suite green (2952 passed locally; the one red is a local SSL fixture that is not checked in).

Branched off `develop` at 3a11555, after #322. It touches `create_redmine_issue`, which #321 and #323 do not, so the only overlap with those is the `### Added` bullet.


jztan:
@andilem merged, thanks. Assigning the staged text to `description` so the custom-field retry sends it too was a nice touch. One request: #324 and #325 have real design questions in them, so please wait for my reply on those before opening PRs. This one was fine because it mirrors #316.

PR #332, opened 2026-09-20
feat(extensions): let a distribution hide its own presentation keys

### LINKED ISSUE #331: unmapped_fields passes through css_classes on every issue, and the size cap cannot catch it
`_issue_unmapped_fields` passes a distribution's own top-level keys through, filtered by size. The comment above the cap names the case it was meant to handle:

> Plugins hang rendering junk off the issue (Easy Redmine's `css_classes`, for one) that is long and of no use to a model. Size is the honest filter here; a per-plugin name list only covers the plugins we happen to have seen.

Measured against a live Easy Redmine, that key is **not long**. It gets through, on every issue, forever.

## Measured

25 issues from one project, `list_redmine_issues(project_id=…, limit=25)`:

| | |
|---|---|
| `css_classes` raw | 99–144 chars, 112 average |
| longest value once wrapped | 217 chars — **22% of `_UNMAPPED_VALUE_MAX_CHARS`** |
| on the wire, per issue | 202 chars |
| one default listing (limit 25) | **5,062 chars** |
| at `limit=100` | **20,248 chars** |

A representative value:

```
issue tracker-203 status-11 priority-24 priority-high2 overdue child
behind-schedule assigned-to-my-group multieditable-container scheme scheme-4
```

That is a CSS class list for Easy's issue grid. Nothing in it answers a question about the issue that the issue's own fields do not answer better — `status-11` next to `"status": {"id": 11, "name": "WAITING"}`, `priority-high2` next to `priority`, `overdue` next to `due_date`.

The size filter cannot reach it. Catching 217 would mean a cap of ~200, which is below a one-paragraph plugin text field — real data would go instead.

## Why a name list is right *here*

The comment's objection is exactly right about a list **this repository curates**: it would cover the plugins you happen to have seen, and go stale. But `ExtensionSpec` already exists so a distribution can state its own facts instead of upstream guessing them — that is what `issue_update_keys` and `issue_query_filters` are. The package that owns `css_classes` knows it is presentation; this repository has no way to know and no business deciding.

So the ask is not a list upstream maintains. It is one more declaration on the spec:

```python
ExtensionSpec(
    family="easy",
    ...,
    issue_payload_skip_keys=("css_classes",),
)
```

merged into `_ISSUE_PAYLOAD_SKIP_KEYS` while the family is enabled, read per call, the same shape as the two seams already there. A server with no extensions behaves exactly as today.

## Deliberately narrow

Only keys a family declares about **itself**. The same instance also sends `is_favorited`, and that one stays: it is per-user state, not presentation, and a caller can reasonably want it. The point is not to trim `unmapped_fields` down to taste — it is that a distribution can name the keys that exist only for its own renderer.

One design question I did not want to decide alone: should `register_extension` refuse a family that declares a skip key another family declares as an update key, or one of Redmine's own names? A skip is a *removal*, so a collision hides data rather than exposing it — the opposite direction from the tool-name checks, but still worth failing startup over, I think. Happy to follow whichever way you prefer.

We run this and would open the PR.

andilem:
Closes #331.

`_issue_unmapped_fields` passes a fork's top-level issue keys through, filtered by size. The comment above the cap names the case it was for — "Easy Redmine's `css_classes`, for one" — and adds that size is the honest filter, because a per-plugin name list would only cover the plugins this repository happens to have seen.

Both halves of that are right, and the key still gets through.

## Measured

One page of 25 issues from a live Easy Redmine:

| | |
|---|---|
| `css_classes` raw | 99–144 chars, 112 average |
| longest value once wrapped | 217 — **22% of `_UNMAPPED_VALUE_MAX_CHARS`** |
| on the wire, per issue | 202 chars |
| one default listing (limit 25) | **5,062** |
| at `limit=100` | **20,248** |

Size cannot reach that. A cap of ~200 would catch it and drop a one-paragraph plugin text field with it. And the value earns none of what it costs:

```
issue tracker-203 status-11 priority-24 priority-high2 overdue child
behind-schedule assigned-to-my-group multieditable-container scheme scheme-4
```

`status-11` sits beside `"status": {"id": 11, "name": "WAITING"}`, `overdue` beside `due_date`, `priority-high2` beside `priority`.

## The seam

What the cap's comment objects to is a list *this repository* curates — and that objection does not apply to a list the extension declares. The package that owns `css_classes` knows it is presentation; nothing here can know it. So it is the same shape `issue_update_keys` and `issue_query_filters` already have, a third time:

```python
    family="acme",
    ...,
    issue_payload_skip_keys=("acme_css_classes",),
))
```

Read per call, gated on the family's `enabled()`, empty and inert on a server with no extensions.

## Checks

Registration refuses a name the serializer emits as a field of its own — that entry never reaches `unmapped_fields`, so it would do nothing, which is worth saying at startup rather than leaving someone to wonder.

Two families naming the same key is **not** an error: both want it gone, and agreeing is not a conflict.

There is deliberately no check in the other direction, and the docs say why: nothing declares the keys a family *reads*, so there is no table to check a skip against, and a guard that checked nothing would only read as though it did. This is the open question from the issue — happy to change it if you would rather have something stricter.

## Tests

Eleven in `TestPayloadSkipKeys`, including the behaviour this changes stated on its own (`test_without_the_seam_it_comes_through`), both serializers, the flag deciding, and the four shape refusals. Full suite green (3005 passed locally; the one red is a local SSL fixture that is not checked in).


jztan (changes requested):
Thanks @andilem. The measurements settle it for me: size can't catch `css_classes`, and a list the extension declares is not the list that comment argued against. I want this in, with two changes first.

1. Check the skip keys against `_ISSUE_PAYLOAD_SKIP_KEYS`, not `_ISSUE_MAPPED_KEYS`. Right now `journals`, `watchers` and `title` register fine and do nothing, which is the case the check exists to report.
2. On your open question: please refuse a skip key that any family declares in `issue_update_keys`, whichever of the two registers first. Those keys are the declared ones, since a caller reads back what it wrote through `unmapped_fields`, and today family A can quietly make family B's attribute write-only. Put it in `_check_issue_seams` with the other cross-family checks, and update the docs paragraph that says there is nothing to check against.

andilem:
Both done in 2527feb.

**1. `_ISSUE_PAYLOAD_SKIP_KEYS`, not `_ISSUE_MAPPED_KEYS`.** You are right that this was the case the check exists to report and it was letting it through. It now compares against the whole set — mapped fields plus the includes, the relations and the search-result keys. The four names from your comment are a parametrize: `subject`, `journals`, `watchers`, `title`.

**2. Skip keys vs `issue_update_keys`.** Your reasoning settles it and I was wrong in the PR body: those keys *are* declared read-back keys, so there is a table to check against, and my "nothing declares the keys a family reads" only held if you ignore the one thing that does. Refused in three directions, so whichever registers first catches it:

- within one spec
- a new skip against a family already writing that key
- a new write against a family already hiding it

Both checks moved into `_check_issue_seams` with the rest. The docs paragraph that claimed there was nothing to check against is gone, replaced by what is actually checked; the `issue_payload_skip_keys` prose now says the refusal out loud, and so does the CHANGELOG entry.

Two families naming the same *skip* key is still allowed — both want it gone.

Eight tests for the two changes, 3012 in the suite.

jztan:
Thanks @andilem, merged. The refusal in `_check_issue_seams` catching all three directions (own spec, skip against an existing write, write against an existing skip) is the shape I wanted, and the measurements in the description made the case for the seam by themselves.

